PR #297, opened 2020-02-19
Click the hash link should get the url to the method

JuanitoFatas:
Right now the `＃` link can‘t be clicked because it has no `href`.

スクリーンショット 2020-02-19 19 30 04

This Pull Request assumes we want people to click `＃` and get the permanent url. Also changed the cursor to pointer so people knows this can be clicked.

### DIFF
--- a/app/views/objects/show.html.slim
+++ b/app/views/objects/show.html.slim
@@ -54,10 +54,10 @@ div class="max-w-screen-xl mx-auto px-3 md:px-0 lg:flex"
                     = seq
           div class="flex md:justify-end w-full md:w-2/12 mt-3 md:mt-0 font-mono"
             - if m.instance_method?
-              a class="px-2 h-6 inline-block rounded bg-gray-200 dark:bg-gray-700 algin-middle cursor-default hover:bg-gray-300 hover:text-gray-800 dark:hover:bg-gray-900 dark:hover:text-gray-400 hover:fill-current" title="Instance Method"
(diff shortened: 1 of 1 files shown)


AlexWayfer:
Good catch! Thank you.

AlexWayfer:
But wait, no, it's not like anchor, it's a sign for instance method. You can see the similar with `::` for class methods, for example: https://rubyapi.org/2.7/o/uri#method-c-decode_www_form_component

And by clicking on method name itself the anchor will be added to the address. But it's not so intuitive, yes.

AlexWayfer:
I guess, it'd be better to see this sign at the left of method name. I'll think and maybe try to change it.

colby-swandale:
@AlexWayfer is correct, the icon is meant to describe a method being either a class or instance method. But i defiantly admit that the icon is not super intuitive.  When i first designed the page, i used a “Instance method” label but i found that to not be good UI. I am defiantly open to any ideas on how to improve it.

@JuanitoFatas I think it will be good to have a link icon show up next to the method names to better signify that it’s a permalink. I can make an issue with an example if you would like?

AlexWayfer:
> @JuanitoFatas I think it will be good to have a link icon show up next to the method names to better signify that it’s a permalink. I can make an issue with an example if you would like?

We can get a link by clicking on method name, do we really need for something else with the same action?

colby-swandale:
Something like what Github does would be great! It's not something entirely detached from the title but adds to the UI to make the it more intuitive to show 
that the method title is a permalink.

Example:
Screen Shot 2020-02-20 at 11 52 50 pm

AlexWayfer:
Yes, maybe. 

Also, maybe the current icon-button of instance/class method can be clickable and do such thing if it'd be placed to the left from method name, as I suggested above. Not sure. Like

> [::] new
> ...
> [#] slice

And `[ ]` clickable.

Issue #289, opened 2020-02-13
Z-index issue between search field and version drop-down


AlexWayfer:
Hello!

There is a small issue with overlapping:

![image](https://user-images.githubusercontent.com/6510020/74429688-77b4e600-4e6c-11ea-969a-5ee3946d94f4.png)

(on the main page)

AlexWayfer:
Thank you!

bronzdoc:
Thank you for the issue report @AlexWayfer !

PR #298, opened 2020-02-20
Fix header navigation opening behind the navbar on the homepage

colby-swandale:
The navigation header doesn't have any z-index ordering set which is causing a problem where navigation elements are currently appearing behind the search bar.

Current Problem:

Untitled -1

This PR fixes the issue by setting the `z-index` accordingly to have the navigation elements showing in the correct order

Untitled

Fixes #298


AlexWayfer:
> Fixes #298

Actually, it fixes #289, #298 is this PR. :grin: 

Typo, I guess.

bronzdoc:
bundlerbot r+

AlexWayfer:
(without change of description you'll have to manually close #289)

PR #312, opened 2020-03-07
Update pagination design

colby-swandale:
The current pagination design is something i neglected when i first added the feature. It's not mobile friendly and the text is too small IMO.

This PR updates the pagination design, which was adapted from https://tailwindui.com/components pagination component.

**Desktop Pagination**

Untitled

**Mobile Pagination**

Untitled 1


AlexWayfer:
Good! But, Mobile pagination look a way less functional then desktop. For example, we can't go to the last page simply. Please, be sure that this is what you want. I don't think that Desktop pagination will be ugly on Mobiles.

colby-swandale:
Thanks for the feedback. When designing the mobile UI i wanted to have a design  similar to sites like Google, and figured most users would only ever visit up to page 2 or 3. I've adjusted the design to not show page 98, 99 etc.

AlexWayfer:
> Thanks for the feedback. When designing the mobile UI i wanted to have a design similar to sites like Google, and figured most users would only ever visit up to page 2 or 3. I've adjusted the design to not show page 98, 99 etc.

For me pagination of Google search results looks so:

image

I'm confused by suggested "Previous" and "Next" buttons — I don't know on what page I'm right now. And how to move to the first one.

Google's pagination does not have such problems.

colby-swandale:
Thanks for the feedback @AlexWayfer. I'm going to merge this now because I really want the better design live now, but I will think about making the mobile pagination better in the future.

PR #318, opened 2020-03-12
Add cache for home page and remove cookies.

ioquatix:


ioquatix:
Consider this a draft until falcon is updated with cache support. However, you may benefit from it via cloudflare's cache. I don't know what other pages could benefit from cache. What controller should we modify next?

ioquatix:
@nateberkopec would love to have your feedback on this, I'm not much of a rails dev.

nateberkopec:
Oh hey @ioquatix for some reason I didn't see your initial ping.

30 minute HTTP cache on most resources seems pretty prudent. Checked prod and it looks like it's working.

ioquatix:
@nateberkopec thanks for checking it in production it's great to have more eyeballs on stuff like this :)

Issue #191, opened 2019-09-26
Allocations are too dam high


colby-swandale:
Not sure why we're allocating 10s of thousands of objects in the object show page.

https://oss.skylight.io/app/applications/k1noEyWLdXuJ/1569499140/45m/endpoints/ObjectsController%23show?mode=memory&responseType=html

natematykiewicz:
Skylight seems fine with the amount of allocations these days. So, I think we're good to close this one!

colby-swandale:
I agree! Thank you again for looking into this! You should totally write up about a blog post about reducing the allocations down by ~50%!

PR #344, opened 2020-03-28
Dramatically reduce allocations

natematykiewicz:
Issue #191 points out that objects#show does _a lot_ of allocations. This PR removes about 32.5k allocations from the the String class page for Ruby 2.7 (https://rubyapi.org/2.7/o/string).

(Note: I did each optimization as a commit. The PR might be easier to review by stepping through he commits.)

# Optimization 1

## Avoid deeply nested hash and array allocations

deep_symbolize_keys returns new objects the entire way through the document tree. Additionally, the HashWithIndifferentAccess.new was an extra allocation per RubyObject and RubyMethod.

The Ruby 2.7 String class has 146 methods, which means 147 allocations just in HashWithIndifferentAccess calls.

Before:
Completed 200 OK in 72ms (Views: 57.9ms | Allocations: 58876)

After:
Completed 200 OK in 51ms (Views: 38.7ms | Allocations: 47559)

Removes ~11,000 allocations

# Optimization 2
## Add frozen_string_literal comment to all applicable files

The magic_frozen_string_literal gem made it very easy to add the "magic comment" to all Ruby files.

Before:
Completed 200 OK in 51ms (Views: 38.7ms | Allocations: 47559)

After:
Completed 200 OK in 54ms (Views: 40.2ms | Allocations: 44923)

Removes ~2,600 allocations

# Optimization 3
## Avoid `url_for` for anchor links
Rails' `url_for` method (which object_path uses) is overkill for simply creating a link to an anchor on the current page.

Each call to `object_path` allocates a Hash, and then the Rails `url_for` logic performs a lot more allocations and router lookups, just to inevitably give us `/2.7/o/string#method-i-chr` for the `/2.7/o/string` page. For an anchor, the `/2.7/o/string` is actually irrelevant. `#method-i-chr` as the href will work just fine.

Additionally, the method names are CGI escaped. A single page will escape the same method name 2-4 times, (depending on if there's both an instance and class variable with the same name). We can remove great deal of allocations by caching the escaped name for the duration of the request.

Before:
Completed 200 OK in 54ms (Views: 40.2ms | Allocations: 44923)

After:
Completed 200 OK in 38ms (Views: 24.8ms | Allocations: 26282)

Removes ~19,000 allocations

# Summary
Before:
Completed 200 OK in 72ms (Views: 57.9ms | Allocations: 58876)

After:
Completed 200 OK in 38ms (Views: 24.8ms | Allocations: 26282)

Removes ~32,600 allocations and 30ms

### DIFF
--- a/app/helpers/search_helper.rb
+++ b/app/helpers/search_helper.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module SearchHelper
   def method_anchor(method)
     # See https://github.com/ruby/rdoc/blob/c64210219ec6c0f447b4c66c2c3556cfe462993f/lib/rdoc/method_attr.rb#L294
--- a/app/helpers/search_helper.rb
+++ b/app/helpers/search_helper.rb
@@ -1,11 +1,15 @@
 # frozen_string_literal: true
 
 module SearchHelper
-  def method_anchor(method)
-    # See https://github.com/ruby/rdoc/blob/c64210219ec6c0f447b4c66c2c3556cfe462993f/lib/rdoc/method_attr.rb#L294
-    method_name = CGI.escape(method.name.gsub("-", "-2D")).tr("%", "-").sub(/^-/, "")
+  def escape_method_name(method_name)
+    (@escaped_method_names ||= {})[method_name] ||= begin
+      # See https://github.com/ruby/rdoc/blob/c64210219ec6c0f447b4c66c2c3556cfe462993f/lib/rdoc/method_attr.rb#L294
+      CGI.escape(method_name.gsub("-", "-2D")).tr("%", "-").sub(/^-/, "")
+    end
+  end
 
-    "method-#{method.instance_method? ? "i" : "c"}-#{method_name}"
+  def method_anchor(method)
+    "method-#{method.instance_method? ? "i" : "c"}-#{escape_method_name(method.name)}"
   end
 
   def result_url(result, version:)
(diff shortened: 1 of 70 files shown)


app/helpers/search_helper.rb
```diff
@@ -1,9 +1,15 @@
+# frozen_string_literal: true
+
 module SearchHelper
-  def method_anchor(method)
-    # See https://github.com/ruby/rdoc/blob/c64210219ec6c0f447b4c66c2c3556cfe462993f/lib/rdoc/method_attr.rb#L294
-    method_name = CGI.escape(method.name.gsub("-", "-2D")).tr("%", "-").sub(/^-/, "")
+  def escape_method_name(method_name)
+    (@escaped_method_names ||= {})[method_name] ||= begin
```
ioquatix: It's not shared between threads is it?

natematykiewicz: I know you already merged this, but no. Each Rails request instantiates a new instance of the view / controller, which the helper methods get mixed into.

I learned a lot from your video about thread safety yesterday. I would have definitely used Concurrent Ruby if it were shared between threads. That’s another optimization that could be made though — because there’s really no point in figuring out the escaped form of the same method name for every Ruby version and class, every single request.

If you’re fine with retaining a little more memory throughout the life of the app, I could make a PR for that change as well!

natematykiewicz:
> Thank you!

You’re welcome!

PR #345, opened 2020-03-28
Cache escaped method names

natematykiewicz:
Cache escaped method names

Every time objects#show is loaded, all of the class and instance
methods are escaped to be used as an anchor link to the method
on the current page.

The thing is, the value a single method name escapes to is constant,
so there's no ned to calculate it every request. Furthermore,
there's a lot of overlap between classes on method names
(how many classes have a `new` method, for example?).

By storing a cache of the html anchors for a method, we can reduce
a significant amount of allocations on the objects#show page.

Additionally, since these are just links to the same page, we can
actually write a manual <a> tag instead of using link_to.

For `/2.7/o/string`, this removes about 7,000 allocations.

Before:
Completed 200 OK in 51ms (Views: 38.9ms | Allocations: 26304)

After:
Completed 200 OK in 39ms (Views: 25.9ms | Allocations: 24993)

Removes ~1,300 allocations

_Note: This cache is shared between threads._

### DIFF
--- a/app/helpers/search_helper.rb
+++ b/app/helpers/search_helper.rb
@@ -1,15 +1,19 @@
 # frozen_string_literal: true
 
 module SearchHelper
+  @@class_method_anchors = Concurrent::Map.new
+  @@instance_method_anchors = Concurrent::Map.new
+
   def escape_method_name(method_name)
-    (@escaped_method_names ||= {})[method_name] ||= begin
-      # See https://github.com/ruby/rdoc/blob/c64210219ec6c0f447b4c66c2c3556cfe462993f/lib/rdoc/method_attr.rb#L294
-      CGI.escape(method_name.gsub("-", "-2D")).tr("%", "-").sub(/^-/, "")
-    end
+    # See https://github.com/ruby/rdoc/blob/c64210219ec6c0f447b4c66c2c3556cfe462993f/lib/rdoc/method_attr.rb#L294
+    CGI.escape(method_name.gsub("-", "-2D")).tr("%", "-").sub(/^-/, "").freeze
   end
 
   def method_anchor(method)
-    "method-#{method.instance_method? ? "i" : "c"}-#{escape_method_name(method.name)}"
+    collection = method.instance_method? ? @@instance_method_anchors : @@class_method_anchors
+    collection.compute_if_absent(method.name) do
+      "method-#{method.instance_method? ? "i" : "c"}-#{escape_method_name(method.name)}"
+    end
   end
 
   def result_url(result, version:)
--- a/Gemfile
+++ b/Gemfile
@@ -27,6 +27,7 @@ group :preload, :default do
   gem "meta-tags"
   gem "sentry-raven"
   gem "rack-attack"
+  gem "concurrent-ruby"
 end
 
 gem "slim"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -375,6 +375,7 @@ DEPENDENCIES
   bootsnap (>= 1.1.0)
   byebug
   capybara (>= 2.15)
+  concurrent-ruby
   dotenv-rails
   elasticsearch-persistence
   falcon


natematykiewicz:
I'm curious what people think of this PR. This is a little more controversial than the previous PR I sent, because it's caching values for the life of the app, across threads.

Also, if you guys do like this idea, what do you think about the cache that I have of `escaped_method_names`?

The pro is that if an instance method and a class method share the name, you only have to do the CGI escape song and dance once instead of twice.
The con is that it uses more RAM for something which may not have much collisions. Of only 5% of method names overlap, the cache probably isn't worth it.

natematykiewicz:
@AlexWayfer I know what you mean, but technically it's new a _new_ dependency. It's just marking it as a dependency of `rubyapi` now.

activesupport, i18n, and sprockets already have it as a dependency, so this doesn't even install a new gem.

If you search the [Gemfile.lock on master](https://github.com/rubyapi/rubyapi/blob/master/Gemfile.lock), you'll see `concurrent-ruby` listed in a few spots.

natematykiewicz:
Alternative idea to this PR:

What if the importer built these values and stored them in the Elasticsearch document? My goal was to avoid doing the same operation over and over and over, for a value that will never change. Perhaps the importer is the better place to do it?

If the importer were updated, you’d need to reindex all of Elasticsearch before updating the view to use it.

If you like that idea, what do you think about indexing the GitHub URL as well? There a ton of allocations to build the URL, because the URI.join instantiates a new URI builder class for each link. That’s another static thing that seems like it could just be prebuilt and be a simple attribute when displaying.

natematykiewicz:
To be clear, https://rubyapi.org/2.7/o/string does 10,000 allocations to build the GitHub links (the `github_url` method in `application_helper.rb`. So of the ~27,000 allocations that happen on the page, nearly half of them (12,500) are something that could just be calculated on import and be regular static attributes when rendering the page.

natematykiewicz:
I just had an idea on how to improve my PR to be a lot simpler.

natematykiewicz:
Okay, I just pushed an update. This is much simpler than what I had originally pushed, and also removes even more allocations. I updated the original post. This removes 7,000 allocations on https://rubyapi.org/2.7/o/string.

Before:
Completed 200 OK in 42ms (Views: 24.7ms | Allocations: 26289)

After:
Completed 200 OK in 38ms (Views: 22.8ms | Allocations: 19365)

app/helpers/search_helper.rb
```diff
@@ -1,15 +1,19 @@
 # frozen_string_literal: true
 
 module SearchHelper
+  @@class_method_anchors = Concurrent::Map.new
+  @@instance_method_anchors = Concurrent::Map.new
```
AlexWayfer: [Class variables are bad](https://rubystyle.guide/#no-class-vars). Can we use class instance variables? With getters, I guess.

ioquatix:
My question about thread safety was to make you think about the life cycle. However, `falcon` spawns all apps in isolated threads/processes, so there is no shared state between instances of falcon. Therefore, you don't need to worry about thread safety issues like this.

natematykiewicz:
@ipquatix, I know you guys bounced around between falcon and puma for a bit (I think you were ironing some kinks with using falcon). Are you saying I should make the assumption that you’ll stay on falcon now? (Considering falcon is yours, I could see that.) I ask, because if you went back to Puma, you’d have a shared cache across all threads.

If you’re for sure sticking with falcon now, then I don’t need to worry about thread safety at all, do I?

Also, am I allowed to assume this is always going to be MRI?

I think what you’re getting at is that I can just use a regular ||= with a Hash, right?

app/helpers/search_helper.rb
natematykiewicz: That comment is in reference to using them in a `class`, where this is a `module`. I personally don't know of a way to use a class variable in a `module`. `SearchHelper` gets included into a dynamically generated renderer class somewhere in Rails internals. I don't have a way to put class methods on that, to my knowledge. To your point, `AutocompleteResult` does also include `SearchHelper`, but is a shared cache not what we want? Why would we want to recompute the value in `AutocompleteResult`?

I personally did intend the cache to be shared across classes, which is one of the few times that a class variable actually is appropriate.

ioquatix:
After thinking about it, and what you said, you should follow the concurrency model dictated by Rails. Because there can be more than one place where threads safety is required, e.g. during testing with Capybara, etc. I don't know all places this could be an issue.

If Rails requires it to be thread safe, you should follow that.

That being said, I hope Rails has good guidelines about this. I'd be interested to know if they specify it clearly or not. Because I have a strong opinion about this: rack apps should not be expected to be thread safe. It's too difficult to know when and where you need to use thread safe primitives.

natematykiewicz:
@ioquatix I'm really grateful for these questions, because it forced me to learn more about how Rails handles thread safety.

After reading https://guides.rubyonrails.org/threading_and_code_execution.html, and digging in the code, it's my understanding that Rails is "thread safe" by default.

You can run multiple requests at once with web servers like Puma or Falcon, but each request is run in isolation via what it calls the ActionDispatch::Executor. (They do similar things for ActiveJob, ActiveRecord, ActionCable, etc.)

https://github.com/rails/rails/blob/1239b22e2250a40ac88709d1bacc904d0f47fafb/actionpack/lib/action_dispatch/middleware/executor.rb#L6

It's a middleware, and is in the default middleware stack.
https://github.com/rails/rails/blob/6cbd81c383f17b2790fbff9d06c091bcadb82ece/railties/lib/rails/application/default_middleware_stack.rb#L42

The app executor is a dynamically generated class that inherits from ActiveSupport::Executor.
https://github.com/rails/rails/blob/6cbd81c383f17b2790fbff9d06c091bcadb82ece/railties/lib/rails/application.rb#L139

Which doesn't do anything useful and just inherits from ActiveSupport::ExecutionWrapper.
https://github.com/rails/rails/blob/6cbd81c383f17b2790fbff9d06c091bcadb82ece/activesupport/lib/active_support/executor.rb

https://github.com/rails/rails/blob/6cbd81c383f17b2790fbff9d06c091bcadb82ece/activesupport/lib/active_support/execution_wrapper.rb

So, it really seems like this ConcurrentRuby stuff I did is unnecessary. A plain ol' Hash seems more than fine.

Do you agree? If so, I can update the PR to simplify this more.

ioquatix:
So you are saying rails will serialise all access to the mutable data structure?

natematykiewicz:
> So you are saying rails will serialise all access to the mutable data structure?

I don't think so.

My understanding of Rails has always been that you're fine to do whatever you want, if it's done using instance methods of ActionController, helpers, and ActiveRecord. Once you start messing around with changing things at the class-level during the life of the request, you need it to be thread safe.

https://guides.rubyonrails.org/threading_and_code_execution.html

> Rails automatically allows various operations to be performed at the same time.
>
> When using a threaded web server, such as the default Puma, multiple HTTP requests will be served simultaneously, with each request provided its own controller instance.
>
> Threaded Active Job adapters, including the built-in Async, will likewise execute several jobs at the same time. Action Cable channels are managed this way too.
>
> These mechanisms all involve multiple threads, each managing work for a unique instance of some object (controller, job, channel), while sharing the global process space (such as classes and their configurations, and global variables). _**As long as your code doesn't modify any of those shared things, it can mostly ignore that other threads exist.**_
>
> The rest of this guide describes the mechanisms Rails uses to make it "mostly ignorable", and how extensions and applications with special needs can use them.

(note the part I highlighted)

It's my understanding that since this PR uses a class variable, we need to be mindful of how the current web server handles concurrency. It's my understanding that if it's a thread-based server (Puma, for example), these class variables will be shared among all threads on the process. If it's a forked-process server (like Unicorn), then obviously, you'll have your own class variables per process.

I don't know where Flacon is at in all of this. 

I know your description of Falcon is:
> Falcon is a multi-process, multi-fiber rack-compatible HTTP server built on top of async, async-io, async-container and async-http. Each request is executed within a lightweight fiber and can block on up-stream requests without stalling the entire server process. Falcon supports HTTP/1 and HTTP/2 natively.

But to be honest, I don't know much about fibers. I don't know how they differ from threads. I know that Falcon uses `Async`, which is a gem to help use fibers, but without understanding what fibers are, I don't know what gotchas may or may not lie in here. To be honest, Falcon's description seems similar to Puma's (except you use fibers and Puma uses threads). Without having an understanding of what a fiber actually is, I can only (possibly wrongly) assume a fiber is similar to a thread, and that the same rules I mentioned earlier regarding Puma would apply.

All in all, I don't know how relevant any of that is.

Since this is MRI, and the cache values don't depend on each other (we aren't trying to increment a counter, where we can have race conditions between threads), I don't know that we really need to worry about thread safety here. I know your video about RubyGems safety this weekend mentioned retrieving partially allocated objects from not reading inside a mutex and having the whole hash become corrupt if its being shuffled around in memory by another thread while you're reading, but now I'm a bit unsure if that's an issue that MRI would actually face, considering the GVL exists.

I personally feel like this change could be reduced down to:

```ruby
@@method_anchors = { true: {}, false: {} }

def escape_method_name(method_name)
  #https://github.com/ruby/rdoc/blob/c64210219ec6c0f447b4c66c2c3556cfe462993f/lib/rdoc/method_attr.rb#L294		    
  CGI.escape(method_name.gsub("-", "-2D")).tr("%", "-").sub(/^-/, "")	
end

def method_anchor(method)
  @@method_anchors[method.instance_method?][method.name] ||= begin
    "method-#{method.instance_method? ? "i" : "c"}-#{escape_method_name(method.name)}"
  end
end
```

Sure, there's the possibility that you get a cache miss on the read and by time you go to write, it's already there. But since the value we're caching is basically constant, I don't personally care. Unless this code leaves the possibility of the whole Hash becoming corrupt (or some other exception occurring), I think it's fine. 1 unnecessary write back into the cache is better than computing the escaped method name every single request.

natematykiewicz:
I've extracted the `link_to` portion of this PR out to #350. At this point, this class variable removes about 1,300 allocations from Ruby 2.7's String page. That's pretty underwhelming to me, for such a controversial change.

@ioquatix I feel like this isn't worth it and we should just close this PR.

I feel like a better PR would be to look into a fragment cache of the view instead of this custom-made caching. That would be more of the "Rails way" of doing it, and would probably be way more performant. I see that the controller seems to try to cache the action for 30 mins. That doesn't seem to be working. On my computer, I'm getting a 200 not a 304, and it takes about 560ms to get https://rubyapi.org/2.7/o/string.

colby-swandale:
I wonder if it would be a good idea to generate the escaped method name when we're processing the ruby-src docs and store it inside ElasticSearch?

natematykiewicz:
@colby-swandale Somewhere in all of my comments on this PR, I suggested that! Obviously, it’d mean reimporting all ruby versions that you’ve already imported. I think that’s the best bet, honestly. Then it’s just a simple attribute read, and super safe.

If you’re up for that, it’d need to be 2 PRs. One would update the importer to store the new field (escaped_name, maybe?), and a second PR to use the new field. You’d need to deploy the importer changes, then reimport everything, THEN deploy the second PR to use the field.

I guess the other option is to do it in one PR with a lazy evaluation.

Add a new :escaped_name in the importer, and also define this in RubyMethod:

```ruby
def escaped_name
  #https://github.com/ruby/rdoc/blob/c64210219ec6c0f447b4c66c2c3556cfe462993f/lib/rdoc/method_attr.rb#L294		    
  body[:escaped_name] || CGI.escape(name.gsub(“-“, “-2D”)).tr(“%”, “-“).sub(/^-/, “”)	
end
```

Then you could later come back and delete the second half of the method once you’ve imported everything.

Thoughts on all that?

headius:
> My question about thread safety was to make you think about the life cycle.

I don't know where this question was so I can't address that, but the code here seems fine from a thread-safety perspective. `Concurrent::Map` is a good choice for the data structure. You may or may not really need `compute_if_absent` since there's no harm in updating the cache twice.

nateberkopec:
> At this point, this class variable removes about 1,300 allocations from Ruby 2.7's String page. That's pretty underwhelming to me, for such a controversial change.

I would agree.

> Once you start messing around with changing things at the class-level during the life of the request, you need it to be thread safe.

This is correct.

> You can run multiple requests at once with web servers like Puma or Falcon, but each request is run in isolation via what it calls the ActionDispatch::Executor.

This is not correct. Executor just helps the Rails framework manage thread-safety internally. It doesn't interact with application code.

>  personally feel like this change could be reduced down to:

IMO, since you have a mutable datastructure being accessed by many threads, you should just use an atomic operation in all cases, regardless of what the GVL will do you for you.

ioquatix:
The other thing to consider, is if you only add to a cache, eventually it can consume a lot of memory that is never released.

natematykiewicz:
Yeah, at some point later in the thread I mentioned that I don’t understand the ActionDispatch::Executor.

I dislike this PR and want to close it.

I think one of 3 things should happen:

1. Use a Rails fragment cache to cache the calculated HTML into Redis, where the cache key is some part of the Elasicsearch document (object name and updated_at sort of thing). Though that would add another network call to request, which may or may not take longer than just recomputing the HTML. This would need to be benchmarked.
2. Update the importer to store this value in Elasticsearch, so this is a simple attribute read. Though, that’ll trade allocations in the helper for still one allocation per method on the model. I think that would reduce about 2,600 allocations to about 180 (if I’m counting correctly, for Ruby 2.7 String class).
3. Not worry about it. I’ve already done a bunch of other improvements to the page, so maybe page is good enough? It seems like the current biggest problem in the page is latency between Elasticsearch and Rails. Also, we aren’t ever 304ing, so sorting that out would improve load speed.

@nateberkopec @ioquatix

ioquatix:
Okay, good discussion everyone, I came out a bit more knowledgeable, so thanks!

headius:
We can throw this into JRuby and gather some performance profiles very easily, if folks are interested. The tools we have can do sampled or instrumented execution profiles as well as allocation profiles with call stack so you can see where they're happening.

ioquatix:
I guess we are all interested in performance and thread safety so I say, personally would love to see some details.

natematykiewicz:
Tonight, I decided to try out storing the `anchor` in Elasticsearch. It's almost the exact same amount of allocations as master. So, it's not worth the trouble, it seems.

natematykiewicz:
Skylight says that it's only like 20ms to render the HTML of the entire page. So there's no point in trying to fragment cache that. It's almost certain that the latency to Redis would be more than that, and 20ms is dang fast.

The latency to Elasticsearch seems to be the biggest problem. At times, it takes over a full second to get a response back.

ioquatix:
Is redis on the same network?

natematykiewicz:
@ioquatix, I'm uncertain. It's set via an ENV.
https://github.com/rubyapi/rubyapi/blob/master/config/environments/production.rb#L49

Grepping around the code, I don't actually see the Rails cache being used. I was hoping I could find a page that used it, so I could see what Skylight had to say about latency.

natematykiewicz:
I made a PR to try out caching the entire page in Redis, skipping Elasticsearch all together. #358

PR #350, opened 2020-04-02
Avoid link_to, to reduce allocations

natematykiewicz:
RE: #191

GET /2.7/o/string

Before:
Completed 200 OK in 33ms (Views: 21.9ms | Allocations: 26279)

After:
Completed 200 OK in 32ms (Views: 19.4ms | Allocations: 20556)

Removes ~5,700 allocations

### DIFF
--- a/app/views/objects/show.html.slim
+++ b/app/views/objects/show.html.slim
@@ -13,7 +13,7 @@ div class="max-w-screen-xl mx-auto px-3 md:px-0 lg:flex"
             ul class="font-mono text-sm"
               - @object.ruby_methods.select(&:class_method?).sort_by(&:name).each do |m|
                 li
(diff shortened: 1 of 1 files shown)


natematykiewicz:
I had never heard of Slim until looking at rubyapi. It seems to really make it easy to avoid the built-in Rails helper methods that tend to allocate a ton of hashes and strings to build out a tag. I still think there's times in this project that `link_to` is valuable, but for these huge loops where hundreds of links are generated, we want to be mindful of whether `link_to` is truly necessary.

colby-swandale:
Wow this is so great! Thank you so much for looking into this!

natematykiewicz:
You’re welcome!

PR #351, opened 2020-04-02
Reduce allocations in github_url

natematykiewicz:
RE: #191

Before:
Completed 200 OK in 54ms (Views: 38.3ms | Allocations: 26401)

After:
Completed 200 OK in 30ms (Views: 18.8ms | Allocations: 19087)

Removes ~7,300 allocations

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module ApplicationHelper
-  GITHUB_REPO = "https://github.com/ruby/ruby/blob/"
+  GITHUB_REPO = "https://github.com/ruby/ruby/blob"
   MASTER = "master"
 
   def master?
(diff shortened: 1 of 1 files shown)


natematykiewicz:
This one might be controversial, but I really think it's a worthwhile change. Typically, I'd be all for the regular `CGI.join` to be safe, but given that all of the parts of the URL are built by the importer, I think it's totally safe to just manually build the URL with slashes ourselves. The `CGI.join` method does a ton of a allocations to build the URL and make sure we aren't doubling up slashes and stuff. The parts of the URL are already file-system safe (they're file paths), because of where they're sourced form, which from what I understand means they're also url-safe.

Additionally, this actually fixes a bug. I was on master with the Array class, and some of the GitHub urls in there don't have a `line`. The old code would have ended the URL with `#L`, where this change will omit the anchor if there is no line.

colby-swandale:
Today i learned! Thank you so much for finding this and making it faster!

natematykiewicz:
You’re welcome!

PR #353, opened 2020-04-03
Add "Parent" and "Included Modules" to sidebar

### LINKED ISSUE #288: `Parent` and `Included Modules` in side menu
Hello!

Please, add `Parent` and `Included Modules` sections in side menu alongside methods.

Reference: https://ruby-doc.org/core-2.7.0/Array.html

---

image

---

image

AlexWayfer:
Resolve #288

### DIFF
--- a/app/models/ruby_object.rb
+++ b/app/models/ruby_object.rb
@@ -51,9 +51,27 @@ class RubyObject
     body[:metadata]
   end
 
+  def ==(other)
+    constant == other.constant
+  end
+
   # This is be empty in search pages
   def ruby_methods
     @ruby_methods ||= body[:methods].collect { |m| RubyMethod.new(m) }
+  end
+
+  def superclass
+    return @superclass if defined?(@superclass)
+
+    value = body[__method__]
+    return unless value
+
+    @superclass = (self.class.new(constant: value) if value)
+  end
+
+  def included_modules
+    @included_modules ||=
+      body[__method__].map { |m| self.class.new(constant: m) }
   end
 
   def to_hash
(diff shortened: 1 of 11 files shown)


app/models/ruby_object.rb
```diff
@@ -56,6 +56,20 @@ def ruby_methods
     @ruby_methods ||= body[:methods].collect { |m| RubyMethod.new(m) }
   end
 
+  def superclass
+    return @superclass if defined?(@superclass)
+
+    value = body[__method__]
```
natematykiewicz, line 62: Is there a reason you used `__method__` here (and in `included_modules`)? This is the first time I've ever seen this. 

Based on the docs I just read, it seems like you could have just used a symbol literal of `:superclass`. Is `__method__` better for some reason?

AlexWayfer: > Is there a reason you used `__method__` here (and in `included_modules`)?

DRY.

> Based on the docs I just read, it seems like you could have just used a symbol literal of `:superclass`. Is `__method__` better for some reason?

Yes, I could. Yes, it's better for DRY reason.

AlexWayfer:
> It seems like there should be some tests for this. Perhaps something in `test/models/ruby_method_test.rb`?

Thank you, added. With new `RubyObject#==` method (I'm not sure about this).

colby-swandale:
This is great! Thank you so much and apologies for taking a while to respond.

natematykiewicz:
I just saw that this got merged and it's really awesome! I've been trying to use Ruby API as my main ruby documentation tool, but that's been hard without this feature (and a few other smaller ones). This is a big deal to me. Thanks!

Issue #354, opened 2020-04-03
Site doesn't ever return a 304


natematykiewicz:
The response headers don't include an `ETag`, so my browser can't send a `If-None-Match`, so that the server can return a 304.

<img width="1792" alt="Screen Shot 2020-04-02 at 10 08 20 PM" src="https://user-images.githubusercontent.com/5104186/78320295-be4ec480-752e-11ea-9c70-40f610dc1722.png">

I thought that the `cf-cache-status: DYNAMIC` seemed suspicious.

https://support.cloudflare.com/hc/en-us/articles/200172516-Which-file-extensions-does-CloudFlare-cache-for-static-content-

> DYNAMIC
> The resource was not cached by default and your current Cloudflare caching configuration doesn't instruct Cloudflare to cache the resource.  Instead, the resource was requested from the origin web server. Use Page Rules to implement custom caching options.

I don't know much about Cloudflare, unfortunately.

natematykiewicz:
@ioquatix do you think this is still a relevant issue?

ioquatix:
An we confirm if-none-match is working?

natematykiewicz:
@ioquatix On localhost, the server returns an ETag, but my browser isn't sending If-None-Match in a subsequent request.

<img width="1180" alt="Screen Shot 2020-04-18 at 9 54 11 PM" src="https://user-images.githubusercontent.com/5104186/79678302-7c28b280-81bf-11ea-8bc4-48536b9a0d5a.png">

Prod does not even return an ETag.

<img width="1180" alt="Screen Shot 2020-04-18 at 9 54 29 PM" src="https://user-images.githubusercontent.com/5104186/79678304-8945a180-81bf-11ea-8d62-29579b94d397.png">

Do we care? I feel like ideally this would work.

ioquatix:
It should work, but maybe cloudflare is doing something weird.

ioquatix:
Soooooo....

```
#!/usr/bin/env ruby

require 'async'
require 'async/http/endpoint'
require 'async/http/client'

Async do |task|
	endpoint = Async::HTTP::Endpoint.parse("https://rubyapi-org.herokuapp.com/2.7/o/string")
	client = Async::HTTP::Client.new(endpoint)
	
	response = client.get(endpoint.path, {
		'if-none-match' => 'W/"578a945b0772ae259625a9e66f06cdff"'
	})
	
	Async.logger.info(response, name: "headers") do |buffer|
[...]

PR #355, opened 2020-04-03
Disable CSRF tokens, site-wide

natematykiewicz:
RE: #354

This site doesn't have a login, so there's nothing to protect.

Also, the random string of the CSRF token in the HTML causes the ETag to never be consistent, since the HTML is ever-changing, which makes the page impossible to cache.

### DIFF
--- a/app/controllers/graphql_controller.rb
+++ b/app/controllers/graphql_controller.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 class GraphqlController < ApplicationController
-  skip_before_action :verify_authenticity_token
-
   def execute
     variables = ensure_hash(params[:variables])
     query = params[:query]
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -31,6 +31,10 @@ Rails.application.configure do
     config.cache_store = :null_store
   end
 
+  # Site doesn't have a login, so the CSRF protection isn't helpful
+  # In fact, it's problematic because the varying CSRF tokens means you can't get a consistent ETag
+  config.action_controller.allow_forgery_protection = false
+
   # Print deprecation notices to the Rails logger.
   config.active_support.deprecation = :log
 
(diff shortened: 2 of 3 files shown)


colby-swandale:
I currently use CSRF to protect the [code execution](https://github.com/rubyapi/rubyapi/blob/master/app/controllers/execute_controller.rb) route, but i think CSRF is entirely broken at the moment because of page caching (it's also not working because the docker server keeps getting locked up)

colby-swandale:
Ah we're caching a CSRF token on the homepage, but the doc pages are not being cached.

Thinking about the CSRF tokens though, I'm not sure it's really protecting anything. If anyone really wanted to use the endpoint, all they would have to do is scrape a CSRF token from the page HTML.

@ioquatix do you have any opinions?

natematykiewicz:
The meta tag is for JS to use. The Rails form helpers inject a hidden field tag in for regular form posts. That reminds me, there probably should be more to this PR, if disabling it is actually the plan.

But the point of CSRF is to prevent someone from doing a destructive action to the server on your behalf. Since there’s no login, I don’t think it’s valuable.

Since the site doesn’t have a login, I don’t think CSRF helps any.

@ioquatix?

ioquatix:
https://github.com/rubyapi/rubyapi/blob/24f8f551e0032b3a4f3b8d6889f2d3a4f8e044a7/app/controllers/application_controller.rb#L46-L49

Is there anything like `skip_csrf_token`?

If not, I agree with this PR.

natematykiewicz:
@ioquatix I've globally disabled CSRF protection in the settings. Now Rails won't inject the 2 meta tags in the `<head>`, or hidden field tags in forms. It also won't attempt to verify them in a `before_action`.

I also removed @colby-swandale manual skipping of verification, since I believe it was strictly there because the CSRF protection feature was broken, and that's the only action that's a non-GET request (meaning it's the only applicable action for CSRF protection).

colby-swandale:
Looks great. Thank you for this!

PR #356, opened 2020-04-03
Importer performance improvement

natematykiewicz:
The `skip_namespace_regex` is constant, so we can save a ton of allocations by making it a constant, instead of building it every time we call `skip_namespace?`, which happens once for every single class and module.

Additionally, `match` creates a MatchData object and creates backrefs (the $~, for example). Since we only care whether it does or doesn't match, we can use the `match?` method, which returns a boolean and doesn't create backrefs.


colby-swandale:
Awesome, thank you!

ioquatix:
Nice improvement!

natematykiewicz:
Thanks @ioquatix

PR #358, opened 2020-04-03
Cache the entire ObjectsController#show page

natematykiewicz:
Avoid even querying Elasticsearch, and just cache the entire HTML string
into Redis.

Before:
Started GET "/2.7/o/string" for 127.0.0.1 at 2020-04-03 01:38:46 -0500
Processing by ObjectsController#show as HTML
  Parameters: {"version"=>"2.7", "object"=>"string"}
ETHON: performed EASY effective_url=http://localhost:9200/ruby_objects_2.7_development/doc/c3RyaW5n response_code=200 return_code=ok total_time=0.007249
  Rendering layouts/application.html.slim
  Rendering objects/show.html.slim within layouts/application
  Rendered objects/show.html.slim within layouts/application (Duration: 2.5ms | Allocations: 3820)
  Rendered layouts/_header.html.slim (Duration: 7.0ms | Allocations: 3454)
  Rendered layouts/application.html.slim (Duration: 12.3ms | Allocations: 8813)
Completed 200 OK in 24ms (Views: 12.7ms | Allocations: 13354)

After:
Started GET "/2.7/o/string" for 127.0.0.1 at 2020-04-03 01:36:17 -0500
Processing by ObjectsController#show as HTML
  Parameters: {"version"=>"2.7", "object"=>"string"}
Completed 200 OK in 1ms (Allocations: 156)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ group :preload, :default do
 
   gem "elasticsearch-persistence"
 
+  gem "actionpack-action_caching"
   gem "http"
   gem "typhoeus"
   gem "kaminari", "~> 1.2.0"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -25,6 +25,8 @@ GEM
       rack-test (>= 0.6.3)
       rails-dom-testing (~> 2.0)
       rails-html-sanitizer (~> 1.0, >= 1.2.0)
+    actionpack-action_caching (1.2.1)
+      actionpack (>= 4.0.0)
     actiontext (6.0.2.2)
       actionpack (= 6.0.2.2)
       activerecord (= 6.0.2.2)
@@ -373,6 +375,7 @@ PLATFORMS
   ruby
 
 DEPENDENCIES
+  actionpack-action_caching
   aws-sdk-s3
   bootsnap (>= 1.1.0)
   byebug
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -5,6 +5,13 @@ class ObjectsController < ApplicationController
 
   before_action :enable_public_cache
 
+  caches_action :show,
+    expires_in: 30.minutes,
+    # Setting the cache path, because I had troubles getting the separate Ruby versions to
+    # have their own caches. Calling `url_for` for /2.7/o/string simply returns /o/string, and
+    # `url_for` is the default cache path.
+    cache_path: -> { "/#{params[:version]}/o/#{params[:object]}" }
+
   def show
     @object = object_repository.find(document_id)
   end


natematykiewicz:
@ioquatix @colby-swandale Check this out. I imagine the Redis responses are really fast. Elasticsearch queries have been known to take upwards of 1.5sec at times (per Skylight). Since this stuff doesn't change long, you really could cache it longer. I just went with 30 mins, since that's what you've got `enable_public_cache` setting the `Cache-Control` header to at the moment.

natematykiewicz:
For background, this `caches_page` was a default feature of Rails. They removed it out to a gem in Rails 4, since the vast majority of a apps have a login, and therefore you can't cache the _**entire**_ page. Because of this, most people use fragment caches, and so the feature was removed from the main library.

It doesn't have a ton of updates, but it's a stable gem that the Rails core team CIs against Rails 6.

ioquatix:
It’s probably not useful once we deploy async-http-cache which will cache everything

natematykiewicz:
Gotcha. I still have no idea what that is. Haha.

Any idea how close you are on that? Do we just close this PR? Or do we merge this and then undo it once you have async-http-cache ready?

ioquatix:
Tonight?

natematykiewicz:
Oh, sweet! Sorry I’m unfamiliar. I didn’t know if it was going to be a few weeks or months or what.

I guess we should just leave this sit here as a fallback plan in case for some weird reason async-http-cache doesn’t work out.

Is there any way you could give me a brief overview of what it does? Does it just cache the response (headers & body) of all GET requests, falling back to Falcon on a cache miss?

ioquatix:
I did a live stream about it. https://www.youtube.com/watch?v=yJWrgXBEYiY

natematykiewicz:
Perfect! Thanks.

nateberkopec:
Yeah, probably easier to let an http cache and the proper cache-control headers so clients/CDNs can do the work for you. _Distributed computing!_ 🎉

ioquatix:
The cache control headers we have now should be okay, no?

nateberkopec:
I can't find a production URL that hits this action, so IDK what they are

ioquatix:
@nateberkopec 

```
% curl -I https://rubyapi.org/2.7/o/string
HTTP/2 200 
date: Sat, 04 Apr 2020 00:25:50 GMT
content-type: text/html; charset=utf-8
set-cookie: __cfduid=d38338ae6ce02806bf8ee6651065730321585959949; expires=Mon, 04-May-20 00:25:49 GMT; path=/; domain=.rubyapi.org; HttpOnly; SameSite=Lax; Secure
x-frame-options: SAMEORIGIN
x-xss-protection: 1; mode=block
x-content-type-options: nosniff
x-download-options: noopen
x-permitted-cross-domain-policies: none
referrer-policy: strict-origin-when-cross-origin
cache-control: max-age=86400, public
content-security-policy: default-src 'self'; img-src 'self'; object-src 'none'; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'
x-request-id: 3735c350-af28-49dc-a71e-d7c38975e219
x-runtime: 0.077971
strict-transport-security: max-age=15552000; includeSubDomains; preload
vary: accept-encoding
via: 1.1 vegur
cf-cache-status: DYNAMIC
expect-ct: max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
server: cloudflare
cf-ray: 57e6e4f63f7fee9e-AKL
alt-svc: h3-27=":443"; ma=86400, h3-25=":443"; ma=86400, h3-24=":443"; ma=86400, h3-23=":443"; ma=86400
```

Specifically: `cache-control: max-age=86400, public`

natematykiewicz:
I know when I looked at the headers yesterday, I noticed that it didn’t return an ETag.


https://github.com/rubyapi/rubyapi/issues/354

ioquatix:
I've increased the duration of the cache to 24 hours. Thanks for the note regarding ETag... I should probably generate it in async-http-cache.

natematykiewicz:
@ioquatix No problem. If you’re planning to cache the entire HTML, then we should do something about CSRF. I suggested disabling it. But caching the entire HTML means everyone has the same CSRF value (at least in the HTML), which I understand is why it was disabled.

https://github.com/rubyapi/rubyapi/pull/355

natematykiewicz:
FWIW, I know Rails uses Rack::Etag. So ETags are made by default by Rails. Not sure if Cloudflare is throwing them away, though.

nateberkopec:
`cf-cache-status: DYNAMIC` indicates that the file isn't cached, probably because the Cloudflare dashboard is set to not cache HTML. Rather than have Falcon do it, why not enable HTML document caching on Cloudflare?

natematykiewicz:
@nateberkopec, how do you invalidate the cache after a deployment? Is there an easy way to do that in Cloudflare?

I don’t know the reasons why @ioquatix has Falcon doing it, but that’s one question I have regarding having Cloudflare do it.

ioquatix:
I don't personally mind where you add caching. However, the application layer cache is something which makes sense to me, even if you have cloudflare doing the caching too.

On cache invalidation, it's not something you can plan on being able to do reliably. Falcon in-memory cache is trashed on deploy/restart so that's predictable. However, because it's a public cache, there is nothing stopping other hops between the server and the user caching the results. It might even be cached in the web browser.

natematykiewicz:
Doesn’t an ETag solve the problem of cache invalidation? Or do I misunderstand what an ETag does?

ioquatix:
ETag does not solve the problem of cache invalidation unless the browser decides to contact the origin server.

natematykiewicz:
Maybe I don't understand how ETag works vs the Cache-Control header. I thought the browser sends a request to the server with the previously retrieved ETag in the If-None-Match header, then the server has to recompute the whole request body (unless it's internally cached the value, of course), and then the server returns a 304 with no body if the values match, or a 200 with a body if they don't. But I know sometimes your browser just returns a cached asset without a web request.

Right now, my browser does a request for the HTML, and then doesn't even do a request for the CSS and JS, because it's cached.

I think our hope is for the browser to do a request for the HTML and either 304 or 200 with new content. But I think a 200 with pre-existing content is not what we want.

Does that seem right?

ioquatix:
> But I know sometimes your browser just returns a cached asset without a web request.

... or local cache, or ISP cache, or some other cache ... there is no guarantee, if you say cache this data for 24 hours, you may never hit the origin server.

My understanding of HTTP caching is a bit limited, but there are some provisions for "use cache if you have it, check with origin too", e.g. `s-max-age`. Haven't fully read up on them so my interpretation is probably wrong. But I think you have to assume it's possible the browser never even checks with the origin server unless the user hits refresh.

There are at least two levels of cache refresh:

- Normal refresh probably does the 304 dance.
- Hard refresh probably sends a request with `cache-control: no-cache`.

natematykiewicz:
@ioquatix I'm getting like 35ms page loads now. I think your work with `async-http-cache` and `async-faraday` made this load way way faster.

I vote that we close this PR.

I'm uncertain if #354 is still relevant. It's still true that the HTML never 304s, but it still loads super fast. So, maybe we don't care?

ioquatix:
I don’t have any strong opinion but if you think the issue is solved that’s awesome glad to hear my work might have had some impact!

ioquatix:
Just for interest what was the latency before?

natematykiewicz:
I remember getting like 300-1000ms regularly.

natematykiewicz:
@ioquatix the `falcon --cache` is caching successfully, right?

ioquatix:
```
koyoko% curl -I https://rubyapi-org.herokuapp.com/2.7/o/string
HTTP/1.1 200 OK
Connection: keep-alive
X-Frame-Options: SAMEORIGIN
X-Xss-Protection: 1; mode=block
X-Content-Type-Options: nosniff
X-Download-Options: noopen
X-Permitted-Cross-Domain-Policies: none
Referrer-Policy: strict-origin-when-cross-origin
Date: Sun, 19 Apr 2020 01:10:25 GMT
Content-Type: text/html; charset=utf-8
Cache-Control: max-age=86400, public
Etag: W/"578a945b0772ae259625a9e66f06cdff"
Content-Security-Policy: default-src 'self'; img-src 'self'; object-src 'none'; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'
X-Request-Id: 7eb1264b-bc90-4d6d-97d0-83d65afe18a6
X-Runtime: 0.057404
Strict-Transport-Security: max-age=31536000; includeSubDomains
Server: falcon/0.36.0
Date: Sun, 19 Apr 2020 01:10:25 GMT
Vary: accept-encoding
X-Cache: hit                         <- Added by async-http-cache
Content-Length: 0
Via: 1.1 vegur
```

ioquatix:
Also FYI cloudflare seems to strip the cache directives but it's also caching at the edge now which is great, GG everyone.

PR #362, opened 2020-04-05
Don't commit macOS .DS_Store files

natematykiewicz:


colby-swandale:
I don't agree in general to adding files into `.gitignore` which don't relate to the project, ie: `.DS_Store`, `Desktop.ini`, etc. I instead recommend configuring your Git environment to ignore these files

See: https://dev.to/johnphamous/tired-of-dsstore-getting-version-controlled-2e0k

natematykiewicz:
Alrighty. I'm used to doing it at work, because otherwise someone always manages to commit one and I feel like I'm having to police the repo to make sure they don't. But, I'll just add it to my own .gitignore file.

PR #363, opened 2020-04-05
Update Gemfile ruby version to match .ruby-version file

natematykiewicz:
To be honest, I'm not sure what this is for, but I know CI (and presumably prod) use the .ruby-version file, which says `2.7.1`.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-ruby "2.7.1"
+ruby "~> 2.7.1"
 
 group :preload, :default do
   gem "rails", "~> 6.0.2"
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-ruby "~> 2.6"
+ruby "2.7.1"
 
 group :preload, :default do
   gem "rails", "~> 6.0.2"


Gemfile
```diff
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-ruby "~> 2.6"
+ruby "2.7.1"
```
colby-swandale, line 3:
```suggestion
ruby "~> 2.7.1"
```

We still probably want to keep the Pessimistic operator so we don't need to update the Gemfile for every ruby 2.7 patch release until 3.0 comes out.

natematykiewicz: That makes sense!

AlexWayfer:
I'd glad to really know for what this is, I'm seeing this thing in the first time.

ioquatix:
`ruby "~> 2.6"` seems okay to me and doesn't conflict with `.ruby-version` because it's upwards compatible.

@AlexWayfer it's used to print warnings if you are using the wrong ruby version, and Heroku uses `.ruby-version` to select which version of Ruby to use.

ioquatix:
Actually maybe I got it backwards, looks like Heroku uses the version from the gemfile?

https://devcenter.heroku.com/articles/ruby-versions

AlexWayfer:
> @AlexWayfer it's used to print warnings if you are using the wrong ruby version, and Heroku uses `.ruby-version` to select which version of Ruby to use.

I see "raise an exception" in [the documentation](https://bundler.io/gemfile.html), and it doesn't look like "print warnings". 😬 But OK, thank you.

But! If environment uses any Ruby version manager (with `.ruby-version` support) — it'll also exit with code 1.

But I believe you that Heroku (somewhy) uses Ruby version from `Gemfile` and not from `.ruby-version`.

natematykiewicz:
https://devcenter.heroku.com/articles/ruby-versions#ruby-version-specifiers

> Bundler locks your Ruby version you are using locally in the Gemfile.lock. In the above scenario if someone with Ruby 2.3.1 on their system runs bundle install then they will get this in the Gemfile.lock:
>
> RUBY VERSION
>    ruby 2.3.1p112


https://github.com/rubyapi/rubyapi/blob/2aa7009a482c23789d1c288a1285e7dd4282b9d0/Gemfile.lock#L410

Sounds like it’s actually using 2.6.5p114, then.

Seems like even more reason to do a hard lock. There seems to be a lot of confusion on which version this actually means. I feel like it’s best to Just write an explicit version and manually update it. Then there isn’t confusion. 🤷🏻‍♂️

AlexWayfer:
> Sounds like it’s actually using 2.6.5p114, then.
> 
> Seems like even more reason to do a hard lock. There seems to be a lot of confusion on which version this actually means. I feel like it’s best to just write an explicit version and manually update it. Then there isn’t confusion. 🤷🏻‍♂️

We have `Gemfile.lock` in VCS, so, if you change something in `Gemfile` (gems or `ruby`) — you should run `bundle` and it should update `Gemfile.lock`, and commit these changes with original changes in `Gemfile`. Please, try.

colby-swandale:
Updated the Gemfile.lock to use the current ruby version and merged into `master`.

Thanks!

PR #375, opened 2020-04-17
DRY home page content

natematykiewicz:
* Make it easier to add / remove tiles while keeping consistency
* Use flexbox instead of adding a `<br>` tag to a specific tile

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -4,5 +4,37 @@ class HomeController < ApplicationController
   before_action :enable_public_cache
 
   def index
+    @tiles = [
+      {
+        class_name: "String",
+        object_path: "string",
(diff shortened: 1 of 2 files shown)


app/helpers/home_helper.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module HomeHelper
+  def home_page_tiles
```
colby-swandale, line 4: Very nitpicky but I think this should be fine to just assign to a `@tiles` var in the controller. Otherwise this looks good! 👍

natematykiewicz:
@colby-swandale D'oh. That's not a nitpick, it's what I should have done in the first place! 🤦‍♂️ 

Done! I also rebased against master.

colby-swandale:
Perfect! Thank you!

natematykiewicz:
No problem!

PR #378, opened 2020-04-18
Remove hover effect from class/instance method indicator

natematykiewicz:
The class/instance method indicators act like a link, but don't actually link anywhere. This actually confused me, because I thought it was supposed to do something.

They still look the same, but the cursor just remains an arrow, and nothing happens on hover.


colby-swandale:
I agree, i think i had an idea when i put them in originally but i can't remember what.

PR #379, opened 2020-04-18
Fix sidebar links

natematykiewicz:
Without running the links through the router, we end up with a path relative to the page we're currently on.

JSON::Ext::Generator::State has a superclass of Object

```ruby
irb(main):006:0> JSON::Ext::Generator::State.superclass
=> Object
```

See the "Parent" on https://rubyapi.org/2.7/o/json/ext/generator/state

The href generated for the Parent was /2.7/o/json/ext/generator/object when it should have been /2.7/o/object


colby-swandale:
Great find! Thank you for the improvement! 🚀

PR #384, opened 2020-04-20
Cache Yarn installation, for faster CI runs

natematykiewicz:
https://github.com/actions/cache/blob/master/examples.md#node---yarn

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -32,12 +32,24 @@ jobs:
           restore-keys: |
             bundler-cache-
 
+      - name: Get Yarn cache directory path
+        id: yarn-cache-dir-path
+        run: echo "::set-output name=dir::$(yarn cache dir)"
+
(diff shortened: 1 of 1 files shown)


natematykiewicz:
I don't have a great way to test this, but I noticed that a recent PR I pushed spent about 28 seconds installing Yarn packages.

Additionally, without the `--frozen-lockfile` option, it was possible to commit changes to the `package.json` file and not a corresponding `yarn.lock` file change. The CI server would `yarn install`, which would generate a dirty `yarn.lock` file, and work using different dependencies than were in the committed `yarn.lock` file. Now, if the `yarn install` attempts to update the `yarn.lock` file, CI will fail.

colby-swandale:
Ah thanks for speeding up CI a ton! Apologies for the delay in my response. 🙇

natematykiewicz:
No problem! Thanks for the merge @colby-swandale.

PR #388, opened 2020-04-21
Upgrade fsevents for macOS Catalina support

natematykiewicz:
https://github.com/fsevents/fsevents/issues/278

fsevents 1.2.9+ is required for proper macOS Catalina support. Without that, a yarn install errors (though it's an optional dependency, so it's not a huge problem). We had 1.2.7 installed.

Yarn doesn't offer a good way to update a subdependency.

https://github.com/yarnpkg/yarn/issues/4986#issuecomment-395036563

This comment suggests deleting the package you want to upgrade from the yarn.lock file and running `yarn install`. So, that's exactly what I did.

Now, fsevents installs without errors.


colby-swandale:
Thanks for resolving this issue! Sorry for taking so long to respond 🙇

PR #445, opened 2020-06-15
Fix incorrect Ruby Version in StandardRB config

natematykiewicz:
RubyApi currently uses Ruby 2.7.1, so 2.6 is wrong.
But, the value defaults to the current ruby version (RUBY_VERSION).

### DIFF
--- a/.standard.yml
+++ b/.standard.yml
@@ -1,7 +1,6 @@
 fix: true
 parallel: true
 format: progress
-ruby_version: 2.6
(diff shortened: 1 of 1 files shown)


colby-swandale:
Thanks!

AlexWayfer:
I suggest to add CI task for this. I've done such things in GitLab CI, but I don't familiar with GitHub Actions.

PR #487, opened 2020-07-24
Fix instance & class methods that share the same name in an object not correctly being imported

### LINKED ISSUE #464: Missing `DateTime` instance methods
Hello.

Here is the `DateTime` documentation: https://rubyapi.org/2.7/o/datetime

image

And here is the same class on Ruby-Doc.org: https://ruby-doc.org/stdlib-2.7.1/libdoc/date/rdoc/DateTime.html

image

And on RubyAPI.org I don't see such methods like `#iso8601`, `#jisx0301`, `#rfc3339`, `#xmlschema`.

colby-swandale:
In the RDoc importer, we perform a small check to ensure we don't import a method with the same name multiple times. This check currently works by just looking at the method name. This check is preventing some legitimate methods that share the same name from being imported because they are different method types.

ie: `DateTime.rfc3339` and `DateTime#rfc3339`

This PR makes a small change to look at both the method name & method type when checking for duplicates.

Resolves #464

### DIFF
--- a/lib/rubyapi_rdoc_generator.rb
+++ b/lib/rubyapi_rdoc_generator.rb
@@ -52,7 +52,7 @@ class RubyAPIRDocGenerator
             depth: constant_depth(doc.full_name)
           }
         }
-        
+
         next if methods.find { |m| m[:name] == method[:name] && m[:method_type] == method[:method_type] }
 
         methods << method
--- a/lib/rubyapi_rdoc_generator.rb
+++ b/lib/rubyapi_rdoc_generator.rb
@@ -53,7 +53,7 @@ class RubyAPIRDocGenerator
           }
         }
 
-        next if methods.find { |m| m[:name] == method[:name] && m[:method_type] == method[:method_type] }
+        next if methods.any? { |m| m[:name] == method[:name] && m[:method_type] == method[:method_type] }
 
         methods << method
       end
--- a/lib/rubyapi_rdoc_generator.rb
+++ b/lib/rubyapi_rdoc_generator.rb
@@ -41,9 +41,7 @@ class RubyAPIRDocGenerator
       methods = []
 
       doc.method_list.each do |method_doc|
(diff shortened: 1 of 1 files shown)


lib/rubyapi_rdoc_generator.rb
```diff
@@ -54,6 +52,10 @@ def generate_objects
             depth: constant_depth(doc.full_name)
           }
         }
+
+        next if methods.find { |m| m[:name] == method[:name] && m[:method_type] == method[:method_type] }
```
natematykiewicz, line 56:
```suggestion
        next if methods.any? { |m| m[:name] == method[:name] && m[:method_type] == method[:method_type] }
```

Since you don't need the return value, it's quite a bit faster to use `any?` (which just returns a boolean), than `find` (which returns the found element).

```ruby
require 'benchmark/ips'

input = 5
vals = 1.upto(10).map { |val| { v: val } }

Benchmark.ips do |x|
  x.report('using #find') do
    vals.find { |val| val[:v] == input }
  end

  x.report('using #any?') do
    vals.any? { |val| val[:v] == input }
  end

  x.compare!
end
```

```
Warming up --------------------------------------
         using #find   191.311k i/100ms
         using #any?   334.715k i/100ms
Calculating -------------------------------------
         using #find      1.927M (± 6.5%) i/s -      9.757M in   5.089815s
         using #any?      3.305M (± 4.3%) i/s -     16.736M in   5.075868s

Comparison:
         using #any?:  3304660.9 i/s
         using #find:  1927283.9 i/s - 1.71x  (± 0.00) slower
```

AlexWayfer:
I've got it! My splitting methods by type in https://github.com/rubyapi/rubyapi/commit/898d1f0a70d0786e809bbd6414ce8d31af570d14 could fix it. Oh…

lib/rubyapi_rdoc_generator.rb
colby-swandale: Ah thanks!

PR #488, opened 2020-07-26
Prevent double-running the test suite on the first commit of a PR

natematykiewicz:
Also, don't run the test suite when a commit is only markdown files.

For example, look at the checks on https://github.com/rubyapi/rubyapi/pull/474
Screen Shot 2020-07-25 at 11 25 00 PM

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,7 +1,21 @@
 name: CI
-on: [push,pull_request]
+
+on:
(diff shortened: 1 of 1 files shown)


natematykiewicz:
After:
Screen Shot 2020-07-25 at 11 26 27 PM

colby-swandale:
Ah thanks!

PR #502, opened 2020-08-12
Method aliases

### LINKED ISSUE #305: Display that methods are aliases
[`Array#size` on rubyapi.org](https://rubyapi.org/2.7/o/array#method-i-size):

image

[`Array#size` on ruby-doc.org](https://ruby-doc.org/core-2.7.0/Array.html#method-i-size):

image

colby-swandale:
This PR adds the functionality to import the alias information from ruby-src/rdoc & display that information to the user where applicable.

Before merging, one question I have is if a method is an alias but also has it's own documentation.  

Resolves #305 

Screen Shot 2020-08-12 at 11 48 39 pm

### DIFF
--- a/app/models/ruby_method.rb
+++ b/app/models/ruby_method.rb
@@ -55,6 +55,22 @@ class RubyMethod
     body[:metadata]
   end
 
+  def alias?
+    method_alias.values.any?
+  end
+
+  def alias_path
+    method_alias[:path]
+  end
+
+  def alias_name
+    method_alias[:name]
+  end
+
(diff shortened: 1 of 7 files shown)


AlexWayfer:
> Before merging, one question I have is if a method is an alias but also has it's own documentation.

Is it resolved?

colby-swandale:
Yea, aliases don't have descriptions

AlexWayfer:
> Yea, aliases don't have descriptions

OK, thanks.

PR #503, opened 2020-08-13
Use method arglists instead of name when a method does not have a documented call sequence

### LINKED ISSUE #492: Method arguments (from stdlib)
Hello.

It'd be nice to see method arguments of stdlib classes.

Example: https://rubyapi.org/2.7/o/fileutils#method-c-mv

image

They are on Ruby-Doc.org: https://ruby-doc.org/stdlib-2.7.1/libdoc/fileutils/rdoc/FileUtils.html#method-c-mv

image

colby-swandale:
Lots of methods currently are only be shown with their name, so no arguments or return information for the user. We ideally want to provide as much information that we have. This PR updates the import process to pull the `arglists` if there is no `call_sequence`.

**Before:**

**After:**

Screen Shot 2020-08-13 at 7 15 16 pm

Resolves #492

### DIFF
--- a/app/views/objects/show.html.slim
+++ b/app/views/objects/show.html.slim
@@ -32,8 +32,10 @@ div class="max-w-screen-xl mx-auto px-3 md:px-0 lg:flex"
                   | #{m.name}
               - else
                 - m.call_sequence.each do |seq|
-                  h4 class="lg:text-2xl text-lg text-gray-900 dark:text-gray-200 font-semibold"
-                    = seq
+                  h4 class="lg:text-2xl text-lg text-gray-900 dark:text-gray-200 font-semibold"= seq
+            - else
+              h4 class="lg:text-2xl text-lg text-gray-900 dark:text-gray-200 font-semibold"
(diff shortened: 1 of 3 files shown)


natematykiewicz:
Ahh, this is why you haven't merged https://github.com/rubyapi/rubyapi/pull/526, isn't it? 😉 

This makes sense to me and I like seeing the arg list. But, can we get some consistency between whether the `h4` or the `a` is on the outside? You probably don't like having the `<a>` on the outside if you can help it, because the `<h4>` is display block and that leaves you with a large click area (which happens already, if there's a call sequence).

colby-swandale:
@natematykiewicz ah sorry it wasn't because of this PR. I'm jumping between 4 different Github projects at the moment and just didn't get to your PR yet.

natematykiewicz:
@colby-swandale I didn't realize this PR was a few weeks old. I got an email when you reopened it, and thought you had just made it.

4 projects sounds like a lot to manage. 😬

PR #527, opened 2020-09-01
Correctly exclude all Bundler, IRB and RDoc classes during import

colby-swandale:
Currently the root `Bundler`, `RDoc`, and `IRB` modules are being imported. We want to exclude these as well.

Screen Shot 2020-09-01 at 11 22 25 pm

### DIFF
--- a/lib/rubyapi_rdoc_generator.rb
+++ b/lib/rubyapi_rdoc_generator.rb
@@ -4,9 +4,9 @@ require_relative "ruby_description_cleaner"
 
 class RubyAPIRDocGenerator
   SKIP_NAMESPACES = [
-    /Bundler::.*/,
-    /RDoc::.*/,
-    /IRB::.*/
+    /Bundler.*/,
+    /RDoc.*/,
+    /IRB.*/
   ].freeze
 
   SKIP_NAMESPACE_REGEX = Regexp.union(SKIP_NAMESPACES).freeze


colby-swandale:
Thanks for the review @natematykiewicz

natematykiewicz:
@colby-swandale I just realized -- are we worried about collisions?

Because if any class uses this in it at all, this new code will omit it. "BundlerFoo"
"MyBundlerClass"
"TheBundler"

I can't think of anything that would matter, but it's worth mentioning that these regexes are very greedy.

Makes me wonder if you shouldn't just take the constant name, split on ::, grab the first element in the array, and see if it's exactly one of these.

```
"Bundler::Foo::Blah".split("::").first
```

natematykiewicz:
Or add anchors to the regexes?

AlexWayfer:
`/Bundler($|::.+)/` is better as well, but we need to check: maybe `split` faster (but requires more objects, `Array`).

colby-swandale:
Sounds good @AlexWayfer ! Did you want to put a quick PR together?

natematykiewicz:
Without a `^` in the front, @AlexWayfer's regex will still match `FooBundler` or `FooBundler::Blah`.

So, you'll want to write some tests for this.

AlexWayfer:
> `/Bundler($|::.+)/` is better as well, but we need to check: maybe `split` faster (but requires more objects, `Array`).

Regexp is faster:

```ruby
# frozen_string_literal: true

# require 'bundler/setup'
# Bundler.setup :system

require 'pry-byebug'

require 'benchmark'
require 'benchmark/ips'
require 'benchmark/memory'


puts '```ruby'
puts File.read(__FILE__).gsub("\t", '  ')
puts '```'
puts
puts '### Output'
puts
puts '```'


EXAMPLES = %w[
  Bundler
  Bundler::Foo
  Bundler::Foo::Bar

  BundlerFoo
  TheBundler
  The::Bundler
  MyBundlerClass
  MyBundler::Class
  My::Bundler::Class
].freeze


VARIATIONS = {
  repexg: (lambda do
    EXAMPLES.map do |example|
      example.match? /^Bundler($|::.+)/
    end
  end),

  split: (lambda do
    EXAMPLES.map do |example|
      example.split('::', 2).first == 'Bundler'
    end
  end)
}.freeze


def test
  results = VARIATIONS.each_with_object([]) do |(name, block), array|
    array << (result = block.call)
    puts "Result of #{name}:"
    pp result
    puts
  end
  uniq_size = results.uniq.size
  puts "Uniq results: #{uniq_size}"
  puts
  exit 1 if uniq_size > 1
end

test

Benchmark.ips do |x|
  VARIATIONS.each do |name, block|
    x.report(name) { block.call }
  end

  x.compare!
end

Benchmark.memory do |x|
  VARIATIONS.each do |name, block|
    x.report(name) { 100.times { block.call } }
  end

  x.compare!
end


puts '```'
```

### Output

```
Result of repexg:
[true, true, true, false, false, false, false, false, false]

Result of split:
[true, true, true, false, false, false, false, false, false]

Uniq results: 1

Warming up --------------------------------------
              repexg    67.734k i/100ms
               split    25.910k i/100ms
Calculating -------------------------------------
              repexg    663.378k (± 3.3%) i/s -      3.319M in   5.008580s
               split    279.710k (± 3.2%) i/s -      1.399M in   5.007408s

Comparison:
              repexg:   663377.7 i/s
               split:   279709.9 i/s - 2.37x  (± 0.00) slower

Calculating -------------------------------------
              repexg    11.200k memsize (     0.000  retained)
                       100.000  objects (     0.000  retained)
                         0.000  strings (     0.000  retained)
               split   103.200k memsize (     0.000  retained)
                         2.400k objects (     0.000  retained)
                        11.000  strings (     0.000  retained)

Comparison:
              repexg:      11200 allocated
               split:     103200 allocated - 9.21x more
```

---



> Sounds good @AlexWayfer ! Did you want to put a quick PR together?

I'm going to create a PR, but I don't know how to better write tests for it, there are even no any tests for generator, sadly.

PR #382, opened 2020-04-19
Split class and instance methods in model

AlexWayfer:
Show them separately on Object page, like on Ruby-Doc.org

Also sort them by name and cache.

Ruby-Doc.org version:

image

`master` version:

image

PR version:

image

PR #534, opened 2020-09-09
Improve methods types styles

AlexWayfer:
Resolve https://github.com/rubyapi/rubyapi/pull/382#issuecomment-689663953

### DIFF
--- a/app/views/objects/show/_methods.html.slim
+++ b/app/views/objects/show/_methods.html.slim
@@ -1,7 +1,7 @@
 div class="md:px-3 my-3"
   h3 class="lg:text-2xl text-lg text-gray-900 dark:text-gray-200 font-semibold mt-8"
     = title
-  hr
+  hr class="mb-6"
 - methods.each do |m|
   div class="md:px-3 my-3"
     a style="display: block; position: relative; top: -80px; visibility: hidden;" id="#{method_anchor m}"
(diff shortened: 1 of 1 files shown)


colby-swandale:
Thank you for working on the PR but don't agree with this change. The method title/call sequence is one of the first things users look at & want to see when arriving on the page, we want these headers to be super easy to distinguish from everything else. Making the headers smaller makes it blend in more easily with method descriptions and harder to distinguish for users.

AlexWayfer:
Small note: I've checked default sizes for HTML headers and compared to Tailwind, there are they:

 HTML | Tailwind
----------|------------
 `<h1>`, `2em` | `.text-3xl`, `1.875rem`
 `<h2>`, `1.5em` | `.text-2xl`, `1.5rem`
 `<h3>`, `1.17em` | `.text-xl`, `1.25rem`
 `<h4>`, `1em` | `.text-lg`, `1.125rem`

So… they're different, but look approximately equal.

Also we have `text-4xl` with `2.25rem` and can use headers starting from it.

natematykiewicz:
@colby-swandale did you see my comment on the "split class and instance methods" that this PR is trying to resolve? I haven't loaded this PR up, so I can't comment on that, but something really should change compared to master.

colby-swandale:
Sorry @natematykiewicz, I did see the comment but didn't get around to replying. I agree that something needs to change between the section titles & method titles, but i am hesitant to change the method title styles as they are - in my mind - one of the most important things to the user on the page.

The solution in my view is to do something similar that I did in https://github.com/rubyapi/rubyapi/pull/382#issuecomment-665610836 and have the section headers styled differently & spaced far between method headers.

colby-swandale:
Sorry I should also say that I am open to other ideas, but I am just hesitant to change the current style of the method headers unless there is a clear benefit for the user.

natematykiewicz:
Can someone post a screenshot of what this PR looks like before and after?

AlexWayfer:
> Can someone post a screenshot of what this PR looks like before and after?

Current `master`:

image

This PR:

image

PR #535, opened 2020-09-09
Improve methods bottom margin

AlexWayfer:
There was a smaller bottom margin for methods with a code block at the end.

Before:

image

After:

image


colby-swandale:
Sorry for the delay in responding. Thanks for the PR.

PR #571, opened 2020-10-07
Add Attributes in object show page

colby-swandale:
This PR adds the final bit of missing information for objects - Attributes, ie `attr_accessor`, `attr_reader`, `attr_writter` to Ruby API. It covers both importing and showing the an Object's attributes on the page in the same way that constants are shown.

Screen Shot 2020-10-07 at 11 27 04 pm


app/models/ruby_attribute.rb
```diff
@@ -0,0 +1,39 @@
+class RubyAttribute
```
natematykiewicz, line 1:
```suggestion
# frozen_string_literal: true

class RubyAttribute
```

app/views/objects/show/_attributes.html.slim
```diff
@@ -0,0 +1,20 @@
...
+            a href="##{a.name}"= a.name
+        div class="flex md:justify-end w-full md:w-2/12 mt-3 md:mt-0 font-mono"
+          span class="px-2 h-6 inline-block rounded bg-gray-200 dark:bg-gray-700 algin-middle cursor-default" title="#{a.access}"
+            = a.access.titleize
+      - if a.description.empty?
+        div class="ruby-documentation py-1"
+          | No documentation available
+      - else
+        div class="ruby-documentation py-1"
+          == a.description
```
natematykiewicz, line 20: You guys recently discussed ending files with new lines. This should have a newline, right?

lib/rubyapi_rdoc_generator.rb
```diff
@@ -86,6 +86,7 @@ def generate_objects
         superclass: superclass,
         included_modules: doc.includes.map(&:name),
         constants: doc.constants.each_with_object([]) { |c, arr| arr << {name: c.name, description: clean_description(doc.full_name, c.description)} },
+        attributes: doc.attributes.each_with_object([]) { |a, arr| arr << {name: a.name, description: clean_description(doc.full_name, a.description), access: readwrite_string(a.rw)} },
```
natematykiewicz, line 89: Any reason these aren't just `map`s?

```suggestion
        constants: doc.constants.map { |c| {name: c.name, description: clean_description(doc.full_name, c.description)} },
        attributes: doc.attributes.map { |a| {name: a.name, description: clean_description(doc.full_name, a.description), access: readwrite_string(a.rw)} },
```

lib/rubyapi_rdoc_generator.rb
```diff
@@ -167,4 +168,15 @@ def format_method_source_body(method_doc)
 
     formatter.format(lexer.lex(method_src))
   end
+
+  def readwrite_string(symbol)
```
natematykiewicz, line 172: Not sure how much it matters (especially since it's just the importer), but a frozen Hash constant would be faster than a case statement. Case statements are O(N), where a Hash lookup is O(1).

Something like:

```ruby
READWRITE_MAPPING = {
  'R' => 'read',
  'W' => 'write',
  'RW' => 'readwrite',
}.freeze

readwrite_string = READWRITE_MAPPING[attribute.rw]
```

natematykiewicz (changes requested):

lib/rubyapi_rdoc_generator.rb
colby-swandale: Ah this is an artefact when I had an alternative design

natematykiewicz (approved):

natematykiewicz:
I like that you're now just storing the text that gets displayed on screen. I also like that you've changed `Read/Write` to `Read & Write`.

AlexWayfer:
For me, it's hard to move eyes from the left (an attribute name) to the right (attribute forms, reader/writer), there are even no lines. 

For example, Ruby-Doc.org displays read-write property right after name:

image

It's easier to read.

I think it's an important information.

colby-swandale:
@AlexWayfer Thanks for the feedback! I see your point about the design being hard to read but i think this issue is a larger design concern. Do you want to open a new issue that we can then explore?

I am going to go ahead and merge this work now so that we can have attributes on the site.

Thanks!

PR #572, opened 2020-10-07
Add OpenSearch description

### LINKED ISSUE #569: Add an Open Search Description for quick-searching from Chrome's Omnibox
First of all, thanks for your work on this project! It's a fantastic resource, much slicker than the official Ruby docs, and I've been using it loads.

It would be really cool if the site had an Open Search Description file (described [here](https://stackoverflow.com/questions/7630144/how-to-add-google-chrome-omnibox-search-support-for-your-site)). Then, you could jump straight into a search on the docs by typing enough of rubyapi.org for Chrome to suggest it, and then hitting Tab to type a search query right in Chrome's URL bar.

I'm happy to take a look at adding this if you think it's a good idea :)

AaronC81:
This PR adds an OpenSearch description XML document, and a `link` to it from the homepage. This means that Chrome automatically detects rubyapi.org as a searchable site!

the Chrome address bar, composing a search on rubyapi.org

Closes #569


app/views/home/index.html.slim
```diff
@@ -13,6 +13,8 @@
           }
         }
 
+  link rel="search" type="application/opensearchdescription+xml" title="Ruby API" href="/opensearch.xml"
```
colby-swandale, line 16: This looks good! My only suggestion is that it would be good to move this into `views/layouts/application.html.slim` so the browser can pickup the OpenSearch configuration for people who are landing on ruby doc pages from Search engines.

AaronC81: I've moved it into there :)

colby-swandale:
looks good! Thanks for the PR!

PR #323, opened 2020-03-15
Darkmode toggle take 2

JuanitoFatas:
## Summary

This Pull Request adds a dark mode switch (html button) in the top right nav menu. The switch defaults to user‘s system darkmode setting. Once toggled will store preference in the `Window.localStorage`.  The switch icons are `☼, U+263C` and `☾, U+263E`.

The switch is managed by a Stimulus controller `theme-switch_controller.js`. When the page loads, we check the OS and localStorage to find user‘s preference and update the switch css class. When user clicks it, we toggle the css class and store preference in localStorage.

When we use [tailwindcss-theming] plugin, the tailwind‘s original colors are all gone. So I added back what we are using in `theme.config.js`, then extract all places using colors in the view into CSS classes.

Added css class for development search query‘s code block display.

Implements #290.

[tailwindcss-theming]: https://github.com/hawezo/tailwindcss-theming

## Try at local

```zsh
$ git clone rubyapi/rubyapi && cd rubyapi
$ git co -b test-pr-323 && git am -3 https://github.com/rubyapi/rubyapi/pull/323
$ bundle install && yarn install && bin/rails s
$ open http://localhost:3000
```

## Review

We could use the css classes separately or extract the common values into generic names. Suggestions welcome.

## Screenshots 

| mobile light | mobile dark |
| ----------- | ------------- |
| スクリーンショット 2020-03-15 23 45 49  | スクリーンショット 2020-03-15 23 45 54 |

Issue #290, opened 2020-02-13
Theme selection


AlexWayfer:
Hello!

It'd be nice to have site UI theme selection somewhere, for example, between light and dark.

colby-swandale:
The day/night themes are currently being set by the [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) media selector. I'm not sure if there is a way to toggle this with JS or if there's a better way we can handle day/night colors with Tailwind CSS. Any help is welcome on this.

JuanitoFatas:
I will look into this.

ioquatix:
I’m okay with automatic theme selection. For me it just works perfectly.

natematykiewicz:
@colby-swandale you can close this now, since #323 was merged, right?

natematykiewicz:
So now that #323 got merged, any chance we could get automatic theme selection added back in? Previously, the site followed the OS setting. Now it's locked to whatever the OS was configured to the first time you visited the site.

I feel like a 3-state option of Light, Dark, Auto (if supported) would be best.

colby-swandale:
Oh you're right @natematykiewicz, thanks for raising that. There should be a third option to use the OS setting. 
 
My immediate thought to resolve this - is to make the button a popover, like the Github and versions popovers and present all 3 options there. Does that sound good to you?

AlexWayfer:
I think, we should have OS-depending theme by default, and save manual change. I think, there is no need for the third option in this case. Scenario:

* You visit RubyAPI with light OS theme — you get light site.
* Then you visit with dark OS theme — you get dark site.
* Then you visit with light OS theme and you get light site again, and so on.
* But, if you once toggle the switcher manually, for example, from light to dark — you now get dark theme all the time, independent from OS theme, it's your choice.

Yes, there may be a wish to reset your choice and come back to auto-mode, but I'm not sure someone will wish so. I suggest to wait for such feedback. 

I think, the most users have single OS theme, light or dark, and switch it very rare. I know about macOS feature with auto-switch, but I think it's rare used too. And if someone uses it — I guess they will not toggle site theme manually.

natematykiewicz:
@AlexWayfer I don't like that at all. A lot of sites offer the 3-state option. Your suggestion means that I could be curious what this new icon does, click it, realize I don't like that, and have no way back because I "made a choice".

[...]

PR #574, opened 2020-10-07
Make server-side aware of dark mode setting

### LINKED ISSUE #573: When dark theme is enabled, light theme flickers visible when loading a page
When dark theme is enabled, each page seems to flicker from light theme to dark theme when first loading:

clicking through several pages on the site, they each load in light theme then flick to dark theme after a couple of seconds

(This is on Chromium 85.0.4183.83 on Arch Linux.)

AaronC81:
This PR switches the dark mode setting to be stored in a cookie rather than in `localStorage`. The server-side can then bake-in the `mode-dark` class if the dark mode cookie is true, removing the flicker of light theme (shown in #573) before the JS runs!

The theme is still checked on page load in JS, in case the theme is changed but the browser has cached the page with the wrong theme or anything like that. It will also move the old `localStorage` dark mode setting into a cookie, if one is found.

Fixes #573.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -25,4 +25,8 @@ module ApplicationHelper
     # Inspiration: https://github.com/rack/rack/pull/1202
     %(#{GITHUB_REPO}/#{version}/#{file}#{"#L#{line}" if line})
   end
+
+  def dark_mode?
+    cookies[:'rubyapi-darkMode'] == "1"
+  end
 end
--- a/app/javascript/controllers/theme-switch_controller.js
+++ b/app/javascript/controllers/theme-switch_controller.js
@@ -5,29 +5,47 @@ export default class extends Controller {
 
   connect() {
     const osDarkMode = window.matchMedia('(prefers-color-scheme: dark)').matches
+
+    // Dark theme was previously stored using localStorage.
+    // Migrate to a cookie if the old localStorage key is present.
     const supportsLocalStorage = 'localStorage' in window
-
     if (supportsLocalStorage) {
-      const darkMode = localStorage.getItem('rubyapi-darkMode')
-
-      if (darkMode !== null && darkMode === '1') {
-        this.setDarkMode(this.switchTarget)
-      } else if (osDarkMode && darkMode === null) {
-        this.setDarkMode(this.switchTarget)
+      const localStorageDarkMode = localStorage.getItem('rubyapi-darkMode')
+      if (localStorageDarkMode !== null) {
+        this.savePreference(localStorageDarkMode)
+        localStorage.removeItem('rubyapi-darkMode')
       }
+    }
+
+    const cookies = document.cookie.split('; ')
+    const darkModeCookie = cookies.find(cookie => cookie.startsWith('rubyapi-darkMode='))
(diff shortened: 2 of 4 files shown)


natematykiewicz:
Pages are cached in CloudFlare. The goal is that the Rails app doesn't receive many request. I'm unsure what Colby will think of this.

natematykiewicz:
In this PR, I fixed a similar issue with FontAwesome. https://github.com/rubyapi/rubyapi/pull/381

I think we could do the same thing -- move the JS out to its own pack that's not deferred and loads before the `application` JS.

AlexWayfer:
This is what I've done in my website: https://github.com/AlexWayfer/alexwayfer.name/commit/65832f4bc563853e869f95f49f0772ab114d8c2b

Summary:
1. CSS class of dark theme for `<html>`.
2. CSS and JavaScript in `<head>`, before `<body>`.
3. JavaScript is on `DOMContentLoaded` event.

natematykiewicz:
The crux of the issue is how long it takes for the `mode-dark` class to be added to the body.

From what I see, Stimulus controllers are executed alphabetically. Renaming `template-switch_controller.js` to something earlier in the alphabet may help. When I worked on the FontAwesome problem, the biggest factor was the `defer` on the `<script>` tag. The JS took a long time to download and execute (which was intentional).

The thing to keep in mind is that some of the JS is progressive enhancement (clicking the search box does something, for example), and some of it is vital to the layout of the page (choosing the correct theme, adding icons to the HTML).

You want the layout stuff to happen immediately, and the progressive enhancements (event listeners) to be deferred.

The CSS downloads and gets processed fast enough. I would personally recommend against inlining any JS / CSS. When JS and CSS are in they own files, their contents can be cached with a really long cache expiration. When it's in the HTML, you end up wasting bandwidth redownloading something that you could have just cached, because the HTML has a much shorter cache duration (because it's volatile due to layout changes of a static URL).

I'm not sure exactly how to solve this problem. If you move the theme switcher JS out of the `application.js` file, then you'll end up loading the Stimulus JS library twice unless you do code splitting.

When the theme switcher was made, I remember wishing there was a way to make it work with normal `prefers-color-scheme: dark` CSS again (like it used to). These CSS properties run much more quickly than deferred JS that adds a class.

Here's the JS in question:
https://github.com/rubyapi/rubyapi/blob/master/app/javascript/controllers/theme-switch_controller.js

Here's the corresponding PR:
https://github.com/rubyapi/rubyapi/pull/323

Here's my comment about wishing there was an "auto" option (though that would only work around this problem):
https://github.com/rubyapi/rubyapi/issues/290#issuecomment-660752580

natematykiewicz:
Screen Shot 2020-10-07 at 11 39 55 PM

The application.js is 3.19MB (609KB Gzipped). I think processing all that JS is going to be slow as well. Why is that file so big? I don't remember it being so big.

AaronC81:
Thanks for your feedback! I didn't consider that there might be a server-side cache.

It looks like this approach isn't going to work out, so I'll close this and give it another shot soon, using the advice in your comments.

natematykiewicz:
Oh man, good call @colby-swandale

PR #577, opened 2020-10-08
Optimize javascript bundle

colby-swandale:
## Context

The current application javascript bundle size is really large (several MBs big) and takes forever for a browser to fully process. This is causing issues with the theme switcher setting the light/dark theme well after the DOM has started rendering and causing a flickering effect.  See #573 

This issue is caused by 2 main issues:

1. when I added the new [repl page](https://rubyapi.org/repl) page, I packaged the entire Monaco editor in the JS application (which is several MBs big).

2. The Monaco editor is packaged in the same javascript bundle that is served for the docs pages.

## Fixes

This PR solves both issues by:

* Configuring the Monaco editor WebPack plugin to only include the bare essentials that's needed for the Repl page.
* Create a separate javascript bundle so that Monaco-editor is only ever loaded when the user visits the Repl page.

## File Before/After

**Before**
Application JS: 3.3MB

** After: **
Application JS: 70KB
Repl JS: 2.2Mb


natematykiewicz (approved):

natematykiewicz:
Once you mentioned that the REPL change made it big, I was thinking extracting the REPL out of application.js would make a ton of sense. This PR looks great to me.

PR #591, opened 2020-10-20
Bulk import objects into ElasticSearch

colby-swandale:


natematykiewicz:
Is this a performance optimization on importing a Ruby version into ElasticSearch?

colby-swandale:
it is

PR #642, opened 2020-12-30
Fixed issue with flickering from light theme to dark on page load

shgtkshruch:
Hi. Thank you for very useful Ruby documentation.

I think the flickering is happening because the theme is switching after the CSS styles are applied. #573 
By separating the theme code from the initial loading script, theme is determined before the styles are applied.

### Before
https://user-images.githubusercontent.com/5207601/103321793-aa64ce80-4a7e-11eb-9700-416c58aee1a2.mov

### After
https://user-images.githubusercontent.com/5207601/103321796-afc21900-4a7e-11eb-9dad-7b8169633484.mov


natematykiewicz:
@shgtkshruch Can you test something for me? I don't think the problem is that the JS was after the CSS. I think the real problem is that the JS adds the theme is deferred, which means it doesn't load until the whole DOM is done (the same as sticking it before the closing `</body>` tag. I think even if you move it after the stylesheet (so, just swapping lines 18 and 19), it'll work fine.

If I'm correct, then I feel like we should have two JS files:

- `essentials.js`:  a synchronous (so just no special attributes to the script tag), which is this set theme code and the icons (and anything else in the future that affects the style of the page).
- `application.js`: a deferred JS file, which contains most things (all the interactions that can wait until after the first paint)

shgtkshruch:
@natematykiewicz thanks comments!
> I think the real problem is that the JS adds the theme is deferred, which means it doesn't load until the whole DOM is done

I think so, too.
As you commented, I created `essentials.js` that include icons and color theme code, and synchronous load by normal script tag.
I test in my local environment and had no problems.

natematykiewicz (approved):
I can't seem to reproduce the problem locally, but when I try out your branch everything still works. This makes sense to me. @colby-swandale

app/javascript/controllers/app/theme-switch_controller.js
```diff
@@ -1,40 +1,43 @@
 import { Controller } from "stimulus"
 
-export default class extends Controller {
-  static targets = ["switch"]
-
-  connect() {
-    const osDarkMode = window.matchMedia('(prefers-color-scheme: dark)').matches
-    const supportsLocalStorage = 'localStorage' in window
+export const setTheme = (switchTarget) => {
```
colby-swandale, line 3: I think it would be beneficial if we keep our JS application code consistent by continuing to use `class` objects rather then exporting a bunch of different functions/constants. What do you think?

shgtkshruch: Thanks for the comment! I agree with that.
I exported `setTheme` separately from `class`, since we only need the code to set the colors when initially loading.
I don't think the "stimulus" at the first load is necessary.
How about split `setTheme` code to new file from this file? (and import it to this file)

colby-swandale:
Hey! I am really sorry for the delay in responding. This looks good and I'm happy for it to be merged in. Thanks!

natematykiewicz:
This definitely seems to have fixed the flickering problems I was seeing!

PR #646, opened 2021-01-01
Update Standard to 0.10.2 and fix offenses

natematykiewicz:


app/services/search/filters/is.rb
```diff
@@ -13,9 +13,9 @@ def self.filter_for(value)
           [{term: {object_type: {value: :class_object}}}]
         when "method"
           [{term: {type: {value: :method}}}]
-        when "class-method", "class-method", "cmethod"
+        when "class-method", "cmethod"
           [{term: {method_type: {value: :class_method}}}]
-        when "instance-method", "instance-method", "imethod", "#"
```
colby-swandale, line 18: These values are not suppose to change, they're keywords used to filter search results.

natematykiewicz: instance-method and class-method are listed twice. It's my understanding that this diff executes exactly the same as the old code. The duplication is why the Deendabot PR failed checks.

colby-swandale: Oh I see, not sure why I did this 🤔

colby-swandale (approved):

app/services/search/filters/is.rb
natematykiewicz: Not sure. I tried following it back in the commits but it looks like you renamed the file at one point. It was there in the original commit of this file.

All I can guess is that you renamed some method in a refactor, and did a find/replace on the method name, which also updated one of the values in the case statement, creating a duplicate.

PR #649, opened 2021-01-01
Update Rails to 6.1

natematykiewicz:
Fix "NoMethodError: undefined method `hour' for 1:Integer" in config/environments/test.rb


colby-swandale:
Thanks!

PR #653, opened 2021-01-08
Upgrade clipboard-copy to 4.0.1 and fix import error

natematykiewicz:
Dependabot's PR https://github.com/rubyapi/rubyapi/pull/613 fails to build with this error:

```
ERROR in ./app/javascript/controllers/app/code_example_controller.js 1:4685-4698

"export 'default' (imported as 'clipboardCopy') was not found in 'clipboard-copy'
 @ ./app/javascript/controllers/app sync _controller\.js$
 @ ./app/javascript/controllers/app/index.js
 @ ./app/javascript/packs/application.js
Child vs/editor/editor:
     2 assets
    Entrypoint main = editor.worker.js editor.worker.js.map
    [5] (webpack)/buildin/global.js 905 bytes {0} [built]
    [6] ./node_modules/monaco-editor/esm/vs/editor/editor.worker.js + 31 modules 452 KiB {0} [built]
        |    32 modules
        + 5 hidden modules
```

I tried `import * as copy from 'clipboard-copy'` and `const copy = require('clipboard-copy')` (the docs tell you to do that `require`). Both of those build fine, but loading the page gives you an error in the console. It varies by browser.

Safari: TypeError: Attempted to assign to readonly property.
Firefox: Uncaught TypeError: "exports" is read-only
Brave: Uncaught TypeError: Cannot assign to read only property 'exports' of object '#<Object>'

It seems to be different ways to say the same thing.

This article mentions setting babel's `sourceType` to `'unambiguous'`. https://medium.com/@po_thiago/cannot-assign-to-read-only-property-exports-of-object-object-1d908589e135

I won't pretend I understand what this is, but it does fix the problem.

The version diff isn't much. I do see that the `clipboardCopy` function is now `async`. Presumably that's what caused the problem. https://github.com/feross/clipboard-copy/compare/v3.1.0...v4.0.1


colby-swandale:
Thanks for investigating this problem and coming up with a good solution! 👏

natematykiewicz:
No problem!

PR #708, opened 2021-03-23
Resolve CI build errors

colby-swandale:


natematykiewicz:
Was the issue a Ruby 3 incompatibility or an Ubuntu incompatibility?

colby-swandale:
I'm guessing was because Github actions changed from [Ubuntu 18.04 to 20.04](https://github.com/actions/virtual-environments/issues/1816) and the FFI gem was looking for an old/supported version of the FFI library.

natematykiewicz:
> I'm guessing this was because Github actions changed from [Ubuntu 18.04 to 20.04](https://github.com/actions/virtual-environments/issues/1816) and the FFI gem was looking for an old/unsupported version of the FFI library. 

Ahh. Makes sense.

natematykiewicz:
@colby-swandale I think it would be wise to not use ubuntu-latest, but instead use whatever Heroku is configured for. Then when a new version comes out on Heroku, you could run the test suite on both OSes, make sure all is well, and then upgrade your Heroku stack and drop the old OS from CI. I'd recommend the same sort of thing with Ruby versions. The matrix is super powerful at avoiding regressions.

HTML-CODDER (commented):
ffi (1.14.2)

PR #716, opened 2021-03-26
Fix breaking bundle install

### LINKED ISSUE #715: Yanked version of `mimemagic` gem is pinned in `Gemfile.lock`
When I run `bundle install` locally, I get this error:

```
Your bundle is locked to mimemagic (0.3.6), but that version could not be found in any of the sources listed in your Gemfile. If you haven't changed sources, that means the
author of mimemagic (0.3.6) has removed it. You'll need to update your bundle to a version other than mimemagic (0.3.6) that hasn't been removed in order to install.
```

It looks like the version of `mimemagic` that we depend on (0.3.6, as of #711) was yanked from rubygems.org.

Rubygem version list: https://rubygems.org/gems/mimemagic/versions
Relevant issue in mimemagic: https://github.com/mimemagicrb/mimemagic/issues/98
Discussion in rails: https://github.com/rails/rails/issues/41750

colby-swandale:
Update the Gemfile to only depend on the specific bits of Rails functionality needed to allow users to `bundle install` again without error.

Resolves #715


natematykiewicz:
@colby-swandale I'm not by a computer at the moment, but usually there's settings in the `config/environments/#{env}.rb` files that need to change too. There's no active storage, active job, or active record settings floating around in there?

natematykiewicz:
@colby-swandale A new Marcel version was released which drops the MimeMagic dependency. Upgrading from Rails 6.1.3 to 6.1.3.1 should fix the issues. 

https://weblog.rubyonrails.org/2021/3/26/marcel-upgrade-releases/

PR #741, opened 2021-04-24
Fix bug where the last character of the search field was being omitted from the autocomplete request

cstyles:
We switched from listening for `keyup` events to `keydown` events in
832b98133b56a00fd940f1c142a15bb0438cab62 because we needed to be able to
call `preventDefault` when the event was an arrow up, arrow down, or
Enter/Return key press. Using `keyup` would mean that we would call
`preventDefault` too late to stop the default behavior.

However, this introduced a bug where the query that we send to the
server is missing its last character. At the time that the `keydown`
action is executed, the input field doesn't yet contain the character
that will be typed when the key goes up.

The easiest way I know of to solve this is to register functions for
both `keyup` and `keydown`. We can handle special characters in
`onKeydown` and normal characters in `onKeyup`.


cstyles:
I feel like this is easiest to see in motion. Here's a clip of the existing behavior:

<details>
<summary>bad clip</summary>

https://user-images.githubusercontent.com/4869194/115965463-3a230280-a4de-11eb-8380-b2e354d5f925.mp4

</details>

Note that the last query is missing the "b" in "Enumerab" and the search results reflect that (they show items for `Enumerator`).

And here's a clip of the behavior on this branch:

<details>
<summary>good clip</summary>

https://user-images.githubusercontent.com/4869194/115965467-4313d400-a4de-11eb-96cc-72dc79904c70.mp4

</details>

Note that the last query _does_ have the final "b" and the search results reflect that.

colby-swandale:
> Using keyup would mean that we would call preventDefault too late to stop the default behavior.

What default behaviour are you referring to here? I tested using `KeyUp` locally and the correct input value was being sent & the keyboard navigation was working correctly.

colby-swandale:
ping @cstyles.

Did you have any thoughts?

cstyles:
> What default behaviour are you referring to here?

For me (using Firefox on macOS), Arrow Up goes to the beginning of the text input field and Arrow Down goes to the end. Also, Enter/Return submits the search form. So without preventing the default behavior, I get something like this:

https://user-images.githubusercontent.com/4869194/116500602-ba8d8e80-a863-11eb-9005-c3a7a165d4c2.mp4

Note that navigating up/down also moves the text cursor. And if I press Enter, it submit the search form instead of selecting the autocomplete suggestion (FWIW, if I tap Enter very quickly, it _does_ select the suggestion. But if I hold it for just a fraction of a second, it submits the search form).

colby-swandale:
I'm happy for this to merge and solve this issue now.

PR #743, opened 2021-04-25
Allow arrow keys to "wrap around" search suggestions

cstyles:
If the user wants to select the last search suggestion, instead of
pressing down a bunch of times, now they can just press up once.

Conversely, if the user has the final search suggestion
selected/highlighted, they can press down to bring themselves back to
the neutral/starting point where no suggestion is selected.


AlexWayfer (changes requested):
I have my vision of this: https://github.com/cstyles/rubyapi/pull/1 😬

colby-swandale:
Thanks!

PR #744, opened 2021-04-25
Fix overzealous regex in `RubyDescriptionCleaner`

cstyles:
The `.*` in these regexes can match more than we expect and cause
problems. Specifically, there's nothing preventing them from comsuming
multiple `<a>` tags which is exactly what happens on some pages (e.g.,
`Array`, `DateTime`, etc.). This commit constrains these regexes so that
they can only match (and thus delete) a single `<a>` tag.

---

Examples of this regex causing problems:

- `Array`
  - https://rubyapi.org/3.0/o/array#class-Array-label-Obtaining+Information+about+an+Array
  - https://ruby-doc.org/core-3.0.0/Array.html#class-Array-label-Obtaining+Information+about+an+Array
- `DateTime`
  - https://rubyapi.org/3.0/o/datetime#class-DateTime-label-When+should+you+use+DateTime+and+when+should+you+use+Time-3F
  - https://ruby-doc.org/stdlib-3.0.1/libdoc/date/rdoc/DateTime.html#class-DateTime-label-When+should+you+use+DateTime+and+when+should+you+use+Time-3F

### DIFF
--- a/lib/ruby_description_cleaner.rb
+++ b/lib/ruby_description_cleaner.rb
@@ -8,8 +8,8 @@ class RubyDescriptionCleaner < Trenni::Sanitize::Filter
     # Most of this method is copy/pasted from Trenni::Sanitize::Filter.parse
     # https://github.com/ioquatix/trenni-sanitize/blob/c8e08d2717b98a2268da89f8196fdb1eb93725bf/lib/trenni/sanitize/filter.rb#L40
 
-    description = description.gsub(/(<a.*&para;<\/a>)/, "")
-      .gsub(/(<a.*&uarr;<\/a>)/, "")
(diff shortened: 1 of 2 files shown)


natematykiewicz (changes requested):
Thanks for the fix!

Please add a test to test/lib/ruby_description_cleaner_test.rb that demonstrates the problem you're solving. That'll ensure that your fix is for sure working, but also that future fixes don't cause a regression.

When I wrote the current tests, I literally copy/pasted descriptions from RDoc and made sure they were output correctly. It would be great if you added an example for this Array problem and/or DateTime problem.

natematykiewicz (approved):

natematykiewicz:
Thanks for the test @cstyles!

natematykiewicz:
@colby-swandale you want to reimport?

PR #770, opened 2021-05-12
Add `remark` lint

AlexWayfer:
For Markdown files.

Main motivation was dollar sign in code blocks which is redundant.

1.  This is well described in the [Markdown Style Guide](https://cirosantilli.com/markdown-style-guide/#dollar-signs-in-shell-code)
    (which default we're using now).
2.  Input symbol is configurable, for example, I have `>`.
3.  Some commands can require root access on some systems,
    like `docker` on Arch Linux without additional actions,
    and there is should be `#`, but it's pretty conditional.


colby-swandale:
> Main motivation was dollar sign in code blocks which is redundant.

Sorry but I disagree with this point, it's very common to use `$` to denote a terminal command. See projects like [aws-cli](https://www.google.com/search?client=safari&rls=en&q=aws+cli&ie=UTF-8&oe=UTF-8), [homebrew](https://brew.sh) etc

AlexWayfer:
> > Main motivation was dollar sign in code blocks which is redundant.
> 
> Sorry but I disagree with this point, it's very common to use `$` to denote a terminal command. See projects like [aws-cli](https://www.google.com/search?client=safari&rls=en&q=aws+cli&ie=UTF-8&oe=UTF-8), [homebrew](https://brew.sh) etc

I've already mentioned why it's not complete correct. Also I forgot about 4th point: GitHub has now "Copy to clipboard" button, and I could not just press it and paste into terminal to start a local copy of RubyAPI, I had to select manually.

This approach is not too popular yet, new shells (like zsh, fish, etc.) are just incoming in our life, as well as "Copy to clipboard" buttons. If you want examples, I made a small research:

* ruby and Rails has `$` in code blocks.
* [Tailwind](https://github.com/tailwindlabs/tailwindcss/blob/master/.github/CONTRIBUTING.md) and [GitHub CLI](https://github.com/cli/cli/blob/trunk/docs/install_linux.md) don't.

Often it correlates with a documentation creation date, old docs remain with `$`.

colby-swandale:
> Often it correlates with a documentation creation date, old docs remain with $.

Again, this seems more like a personal take than something that's now a general assumption to most developers, but I'm happy to remove the `$` anyway so that "Copy to Clipboard" works.

Going through the PR, I struggle to see the need to introduce a markdown linter. If there was a blog and we were writing lots of markdown files from different contributors, I would certainly see the usefulness of having one, but this project only has 3 Markdown files which rarely change.

AlexWayfer:
> Going through the PR, I struggle to see the need to introduce a markdown linter. If there was a blog and we were writing lots of markdown files from different contributors, I would certainly see the usefulness of having one, but this project only has 3 Markdown files which rarely change.

But when it changes — it's better to be in the same rules as existing content.

After #776, which I highly support, I can introduce a new job with files change check. I don't think it'll complicate somehow significantly a dev workflow.

.github/workflows/main.yml
```diff
@@ -4,20 +4,53 @@ on:
...
   BUNDLE_PATH: vendor/bundle
 
 jobs:
-  lint:
+  lint-markdown:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: dorny/paths-filter@v2
```
AlexWayfer, line 21: I saw somewhere a syntax for semi-native GH Actions conditions, but lost it and didn't find.

.github/workflows/main.yml
```diff
@@ -4,20 +4,53 @@ on:
...
     branches:
       - main
-    paths-ignore:
-      - '**.md'
   pull_request:
     types:
       - opened
       - synchronize
-    paths-ignore:
-      - '**.md'
```
AlexWayfer, line 14: We can use conditions for Ruby jobs later, but there is no simple way with current GH Actions syntax:
* As I understand, `if changed_files != 'markdown'` will skip Ruby job for commits with Ruby files **and** Markdown (or other non-related) files.
* But enumeration of all related to Ruby jobs files can be huge.

colby-swandale:
@AlexWayfer 👋🏻 I have looked at this PR again and changed my mind since my last comment. I would like to see this change merged. If you can rebase this branch, I will work on getting this merged. Thank you.

AlexWayfer:
> @AlexWayfer 👋🏻 I have looked at this PR again and changed my mind since my last comment. I would like to see this change merged. If you can rebase this branch, I will work on getting this merged. Thank you.

Great to hear it. I'll try after (new) basic issues like #1139 (don't hurry up anyone).

PR #788, opened 2021-05-27
Update setup script & Update README to use setup script to automate Bundler and Yarn installs

colby-swandale:
Small quality of life improvement to make setup and getting started easy.


natematykiewicz:
@colby-swandale any reason to not merge this? I tried it and it works great!

PR #857, opened 2021-08-18
Don't link to objects in the sidebar which do not exist

### LINKED ISSUE #481: Tempfile parent class 404
https://rubyapi.org/2.7/o/tempfile

Screen Shot 2020-07-21 at 9 40 42 AM

The parent class there links to https://rubyapi.org/2.7/o/delegateclass(file), which 404s.

The ruby-doc page for Temple also lists `DelegateClass(File)` as the parent class, but it's not a link.
https://ruby-doc.org/stdlib-2.7.1/libdoc/tempfile/rdoc/Tempfile.html

Screen Shot 2020-07-21 at 9 42 16 AM

colby-swandale:
We shouldn't create links to objects which the documentation refers to, but we don't have a page for

Resolves #481

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -45,6 +45,14 @@ class ApplicationController < ActionController::Base
   end
   helper_method :home_path
 
+  def object_exists?(klass)
+    path = klass.is_a?(RubyObject) ? klass.id : RubyObject.id_from_path(klass)
+    !!RubyObjectRepository.repository_for_version(ruby_version).find(path)
+  rescue Elasticsearch::Persistence::Repository::DocumentNotFound
+    false
+  end
+  helper_method :object_exists?
+
   def skip_session
     request.session_options[:skip] = true
   end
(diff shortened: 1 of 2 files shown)


natematykiewicz:
What’s the perf on this like? Because it looks like it’s going to do N queries. I guess since it’s just the modules/classes it won’t be that many queries…

natematykiewicz:
It looks like this is loading up the entire documentation for the parent class and each included module.

So viewing https://rubyapi.org/3.0/o/stringio, will ask ElasticSearch for the documentation of the entire `StringIO`, `Object`, `Enumerable`, `IO::generic_readable`, and `IO::generic_writable` objects.

I'm not very familiar with ElasticSearch, but if I were doing this in SQL, I'd do something like this:

```ruby
objects = [@object.superclass] + @object.included_modules
object_names = objects.compact.map(&:name)

existent_object_names = `SELECT name FROM ruby_objects WHERE ruby_version = ? AND name IN (?)`
```

Do that query once for the request, and then you can simply check whether the superclass or included module's name is in the list of ones that actually exist in the database.

PR #997, opened 2022-01-23
drop webpacker for esbuild

colby-swandale:
This PR replaces Webpacker with esbuild. The changes follow the conventions established by rails in the [js-bundling repo](https://github.com/rails/jsbundling-rails), but some additional changes were required.

* Replace the `application.sass` with `application.scss` as esbuild does not have native support. Do we really care?
* Upgrade to tailwindcss (needed for the css-bundling gem), luckily there are no breaking changes specifically our uses.


colby-swandale:
@natematykiewicz I think I've now finished with this PR. Would you be able to test this PR on your dev machine and let me know if you run into any issues?

natematykiewicz:
Looks good to me! I followed the updated instructions in the Readme. Everything seems to look identical to prod. I can barely tell you changed anything. The only difference I see is that the console doesn't have big Webpack outputs anymore, and the JS compiles faster. So, awesome!

natematykiewicz:
One cool thing I've seen people do with `bin/dev` is check to see if [Overmind](https://github.com/DarthSim/overmind) is installed, and if it is, prefer it over Foreman.

colby-swandale:
Yea sounds good! Happy to merge a PR that adds that.

PR #1024, opened 2022-02-24
Fix: github url is invalid on dev & 3.1 version

xyzhou-1:
The `ruby_31` version is parsed as '3.1' which has no patch part('3.1.1'),that causes the `github_url` result is invalid(https://github.com/ruby/ruby/blob/v3_1/object.c).

The same problem happens on dev version,because there is no tag or branch named `dev`.

This pr lets the `View on GitHub` guide to valid url on dev & 3.1 version.
image

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -16,11 +16,14 @@ module ApplicationHelper
   # Map a method source file into a url to Github.com
   def github_url(ruby_doc)
     version, file, line = ruby_doc.source_location.split(":")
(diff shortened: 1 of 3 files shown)


colby-swandale:
Apologies for the delay in response. This looks good!

Thank you ! 🙇🏻

natematykiewicz:
@colby-swandale do you need to do anything to reimport the RDoc data? Or is this all just HTTP cached for a bit?

colby-swandale:
I have run a reimport that should fix the URLs now

PR #1110, opened 2022-06-06
Type Signatures

colby-swandale:
This PR adds a new feature to import the type signatures of the ruby-src, using the RBS gem and allows the user can optionally toggle that on the Ruby Object page.

### DIFF
--- a/app/models/ruby_method.rb
+++ b/app/models/ruby_method.rb
@@ -23,6 +23,10 @@ class RubyMethod
     body[:object_constant]
   end
 
+  def signitures
+    body[:signitures]
+  end
+
   def class_method?
     method_type == "class_method"
   end
@@ -104,7 +108,8 @@ class RubyMethod
       call_sequence: call_sequence,
       alias: method_alias,
       source_body: source_body,
-      metadata: metadata
+      metadata: metadata,
+      signitures: signitures
     }
   end
 
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -23,7 +23,9 @@ Rails.application.configure do
     config.action_controller.perform_caching = true
     config.action_controller.enable_fragment_cache_logging = true
 
-    config.cache_store = :memory_store
+    config.cache_store = :redis_cache_store, 
+      url: ENV.fetch("REDIS_CACHE_URL") { "redis://localhost:6379/1" }
+
     config.public_file_server.headers = {
       "Cache-Control" => "public, max-age=#{2.days.to_i}"
     }
@@ -60,4 +62,8 @@ Rails.application.configure do
   # Site doesn't have a login, so the CSRF protection isn't helpful
   # In fact, it's problematic because the varying CSRF tokens means you can't get a consistent ETag
   config.action_controller.allow_forgery_protection = false
+
+  config.session_store :cache_store, key: "_sessions_development",
+    compress: true, pool_size: 5, expire_after: 1.month
+    url: ENV.fetch("REDIS_SESSION_URL") { "redis://localhost:6380/1" } 
 end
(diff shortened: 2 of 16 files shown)


app/models/ruby_method.rb
```diff
@@ -23,6 +23,10 @@ def object_constant
     body[:object_constant]
   end
 
+  def signitures
+    body[:signitures]
```
natematykiewicz, line 27: Signatures actually has an "a", not "i". Only commenting on this one, though there's a few other spots in this PR with the same typo.

```suggestion
  def signatures
    body[:signatures]
```

config/environments/development.rb
```diff
@@ -60,4 +62,8 @@
   # Site doesn't have a login, so the CSRF protection isn't helpful
   # In fact, it's problematic because the varying CSRF tokens means you can't get a consistent ETag
   config.action_controller.allow_forgery_protection = false
+
+  config.session_store :cache_store, key: "_sessions_development",
+    compress: true, pool_size: 5, expire_after: 1.month
```
natematykiewicz, line 67: You're missing a comma at the end of this line, so the `url` is likely just a hash literal that's not doing anything. Though maybe the `url` line just needs to be deleted. The same thing is in the production file.

```suggestion
    compress: true, pool_size: 5, expire_after: 1.month,
```

config/environments/production.rb
```diff
@@ -87,4 +88,8 @@
 
   # Disable CSRF protections
   config.action_controller.allow_forgery_protection = false
+
+  config.session_store :cache_store, key: "_sessions_production",
+    compress: true, pool_size: 5, expire_after: 1.month
+    url: ENV.fetch("REDIS_SESSION_URL") { "redis://localhost:6380/1" }
```
natematykiewicz, line 94: See comment in `config/environments/development.rb`

natematykiewicz (changes requested):

lib/ruby_type_signiture_repository.rb
```diff
@@ -0,0 +1,70 @@
...
+      return lookup_type_method_signiture(object_type, method, method_type: :class)
+    end
+  end
+
+  private
+
+  def rbs_type_for_object(klass, &block)
+    type = RBS::TypeName.new(name: klass.to_sym, namespace: RBS::Namespace.root)
+    yield type if block_given?
+    type
```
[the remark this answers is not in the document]
colby-swandale: Ummmmm, good question, I wrote this code like 1.5 years ago, and i can't exactly remember why now 🤔

app/models/ruby_method.rb
colby-swandale: Was in the middle of fixing this but got distracted. Thanks for reminding me.

lib/ruby_type_signiture_repository.rb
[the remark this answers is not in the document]
colby-swandale: I've updated the import to not use blocks 👍🏻

natematykiewicz:
Sounds good @colby-swandale! I see a few more `signiture` typoes in lib/ruby_type_signature_repository.rb. And I'm getting CSP errors. I know these pages are cached in Cloudflare, but I do see it live on https://rubyapi.org/3.1/o/bigdecimal.

Screen Shot 2022-06-13 at 10 49 36 AM

colby-swandale:
I have purged the Cloudflare cache, so these errors should disappear now.

natematykiewicz:
Still getting those CSP errors.

colby-swandale:
Thanks! @natematykiewicz. I'll look into what's happening.

natematykiewicz:
Thanks @colby-swandale! I was experiencing this in Safari, by the way.

natematykiewicz:
Here's what I see in Firefox:
Screen Shot 2022-06-13 at 10 37 39 PM

PR #1112, opened 2022-06-09
Move Ruby Version config to anyway_config

colby-swandale:
The current ruby configuration is a painpoint. We have to maintain different configuration options to indicate with versions of Ruby are EOL, which version is the "default", and perform calculations to know which set is currently being maintained by the core team.

This PR introduces the gem [anyway_config](https://github.com/palkan/anyway_config) that moves this configuration into a single YML file that we can compile into a RubyVersion that we easily filter and lookup.

### DIFF
--- a/config/ruby.yml
+++ b/config/ruby.yml
@@ -1,0 +1,23 @@
+default: &default
+  versions:
+    - version: 3.1
+      default: true
+    - version: 3.0
+    - version: 2.7
+    - version: 2.6
+    - version: 2.5
+      eol: true
+    - version: 2.4
+      eol: true
+    - version: 2.3
+      eol: true
+    - version: dev
+
+development:
+  <<: *default
+
+test:
+  <<: *default
+
+production:
+  <<: *default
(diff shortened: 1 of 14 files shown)


config/ruby.yml
```diff
@@ -0,0 +1,23 @@
+default: &default
+  versions:
+    - version: 3.1
+      default: true
```
AlexWayfer, line 4: I'd take `default` out of `versions` to escape multiple `default` (occasionally).

And probably add a check to including `default` in `versions`.

natematykiewicz: Is there any reason to not just have the first version in the list be the default?

AlexWayfer: > Is there any reason to not just have the first version in the list be the default?

Not sure. Probably if we'll have some beta or `head` version docs. But until not — it's an option too.

natematykiewicz: There already is. It's called "dev" and it's the last version in this YAML file.

AlexWayfer: > There already is. It's called "dev" and it's the last version in this YAML file.

Oh, I'm sorry. In this case, I'm not sure that we should place it as the latest one (excluding obsolete). Like… it's obviously more than latest (3.1 for now), but it's under 2.6 and above 2.5.

PR #1118, opened 2022-06-14
Prefer overmind, if it's installed

natematykiewicz:
[Overmind](https://evilmartians.com/chronicles/introducing-overmind-and-hivemind) is much better to work with than Foreman because it actually lets you interact with debuggers. For this reason, it's nice to use overmind if it's installed.


colby-swandale:
Sounds good!

PR #1122, opened 2022-06-16
Move Type Signature toggle to Controller Action

colby-swandale:
I recently added a new feature to show RBS Type Signatures if a user clicks on a button. See #1110. This feature uses  Stimulus Reflex, which uses WebSockets to update the page with the signatures/call sequences. Unfortunately, toggling this feature takes a long time to process, about 10 seconds which is undesirable. I'm not sure why this is.

This PR changes the implementation of the feature to use a simple controller action to toggle the feature on/off. Unfotunently I lose the smooth Javascript-based transition and have to rely on redirects instead, but I think this trade-off is okay overall.


colby-swandale:
@natematykiewicz, thanks for the feedback. I initially implemented this idea in Stimulus but found it became very complicated. There was too much unnecessary state management, which just became a pain to wrap my head around. Stimulus Reflex was a way to reduce that complexity to just a couple of lines with a session cookie, which works very well (minus the performance problem). I prefer to keep this feature simple, which, frankly, I'm more than happy to trade off for Cloudflare cache.

natematykiewicz:
Yeah I'm not sure how it works in Stimulus. I've not used it a ton. The old school jQueryist in me would just make 2 classes `.type-signature` and `.call-stack` and wrap the code in the `if` and `else` in divs with those classes. Then on button press I'd do check to see if the local storage key is set. And add/remove the `hidden` class from all of the `call-stack` and `type-signature` elements, update the text, and local storage.

PR #1123, opened 2022-06-16
move ElasticSearch config into elasticsearch.yml using anyway_config

colby-swandale:


natematykiewicz:
You're turning me into a fan of Anyway Config. I'd not seen it before.

AlexWayfer:
> You're turning me into a fan of Anyway Config. I'd not seen it before.

I always talk that file configs (with examples) are better than flat ENV variables! Even through classes.

PR #1127, opened 2022-06-19
Move Repl config into an AnywayConfig object

colby-swandale:


config/configs/repl_config.rb
```diff
@@ -0,0 +1,3 @@
+class ReplConfig < ApplicationConfig
+  attr_config url: "http://localhost:3001/", api_key: ""
+end
```
AlexWayfer, line 3: There is [EditorConfig](https://github.com/rubyapi/rubyapi/blob/1c207e0/.editorconfig#L9). :cry:

PR #1139, opened 2022-06-22
Change default ports for Redis, PostgreSQL and ElasticSearch

AlexWayfer:
They already can be run in background, so `docker compose up` can fail.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -43,7 +43,7 @@ Rails.application.configure do
 
   # Use a different cache store in production.
   config.cache_store = :redis_cache_store, {
-    url: ENV.fetch("REDIS_CACHE_URL") { "redis://localhost:6379/1" },
+    url: ENV.fetch("REDIS_CACHE_URL") { "redis://localhost:6381/1" },
     ssl_params: {
       verify_mode: OpenSSL::SSL::VERIFY_NONE
     }
@@ -101,7 +101,7 @@ Rails.application.configure do
       pool_size: 5,
       expire_after: 1.month,
       url: ENV.fetch("REDIS_SESSION_URL") { "redis://localhost:6380/1" },
-      ssl_params: { 
+      ssl_params: {
         verify_mode: OpenSSL::SSL::VERIFY_NONE
       }
     }
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -27,7 +27,7 @@ jobs:
         uses: ruby/setup-ruby@v1
         with:
           bundler-cache: true
-      
+
       - name: Run linting
         run: ./bin/standardrb --no-fix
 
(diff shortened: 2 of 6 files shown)


config/environments/production.rb
```diff
@@ -43,7 +43,7 @@
 
   # Use a different cache store in production.
   config.cache_store = :redis_cache_store, {
-    url: ENV.fetch("REDIS_CACHE_URL") { "redis://localhost:6379/1" },
+    url: ENV.fetch("REDIS_CACHE_URL") { "redis://localhost:6381/1" },
```
natematykiewicz, line 46: I don't think you want to change production ports

natematykiewicz:
> Probably we should not expose dependencies ports at all?

What do you mean?

AlexWayfer:
> > Probably we should not expose dependencies ports at all?
> 
> What do you mean?

As I know, we can delete these ports exposing and they (services) will be available under nicknames let's say (like `redis:6379`, etc.)

AlexWayfer:
But… as I see, the web application is not in Docker (including compose). I think we should move at this direction. But I know 0 about production.

And, at this moment. I'm unable to work on RubyAPI just because it's trying to expose (default) ports which are already bind (by system packages).

AlexWayfer:
Heroku can deal with docker-compose: https://devcenter.heroku.com/articles/local-development-with-docker-compose

Like, we need for a service like `web` there.

I'll wait for some comments from @colby-swandale, no problem.

colby-swandale:
👋🏻 Docker was mostly just a personal preference when I started the project; it's easy to document & start/reset with a few simple commands. Though Docker is the "blessed" method that Ruby API will work with out of the box, If you prefer your OS package manager, that's perfectly fine and should work without issue. I like happy developers over enforcing ways of working.

I don't desire a `web` service in Docker Compose. I always run rails/node/python applications outside Docker, so adding that service will be disruptive when I run `docker compose up`. I will happily support a [docker-compose.overwrite.yml](https://docs.docker.com/compose/extends/) file with a web service for people who prefer to run Rails inside Docker.

AlexWayfer:
> I don't desire a `web` service in Docker Compose. I always run rails/node/python applications outside Docker, so adding that service will be disruptive when I run `docker compose up`.

May I know why? Just curious.



> I will happily support a [docker-compose.web.yml](https://docs.docker.com/compose/extends/) file with a web service for people who prefer to run Rails inside Docker.

Sounds like not a difficult change, I'll try.

AlexWayfer:
> > I will happily support a [docker-compose.web.yml](https://docs.docker.com/compose/extends/) file with a web service for people who prefer to run Rails inside Docker.
> 
> Sounds like not a difficult change, I'll try.

Update: it'll require changes like this:

```diff
-  url: <%= ENV.fetch("REDIS_SESSION_URL") { "redis://localhost:6379/1" } %>
+  url: <%= ENV.fetch("REDIS_SESSION_URL") { "redis://redis:6379/1" } %>
```

And I don't think it'll work for you.

colby-swandale:
I'm closing this PR for now, but feel free to open a new PR with rebase if you would like to see this merged in the future.

PR #1144, opened 2022-06-29
Add a "System Default" option to the theme switcher

natematykiewicz:
I wanted an option to choose "System Default".

I don't love the icon. I used `adjustments` from https://heroicons.com. I feel like it needs to be a dropdown menu like the version switcher. If anyone wants to improve this, please do.

Right now, clicking the version switcher toggles between auto, light, dark, auto, light, dark.

Screen Recording 2022-06-29 at 1 13 16 AM

### DIFF
--- a/app/javascript/controllers/app/theme-switch_controller.js
+++ b/app/javascript/controllers/app/theme-switch_controller.js
@@ -10,41 +10,50 @@ const moonIcon = `
   <path d="M17.293 13.293A8 8 0 016.707 2.707a8.001 8.001 0 1010.586 10.586z" />
 </svg>`
 
+const systemDefaultIcon = `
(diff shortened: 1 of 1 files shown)


colby-swandale:
> I don't love the icon. I used adjustments from https://heroicons.com/. It needs to be a dropdown menu like the version switcher. If anyone wants to improve this, please do.

I agree. I would love to see this added, but I feel the icon doesn't translate into the action (i think of an optional menu/modal showing up) & a single button being a toggle for three different states is not ideal. I experimented a few months ago using the [stimulus-dropdown](https://www.stimulus-components.com/docs/stimulus-dropdown/) library to show a list of the themes (including a system option), but it depends on stimulus reflex. We should use that but reuse the existing JS code to switch the themes.

AlexWayfer:
I agree that drop-down is a good option.

Do we use only heroicons or we have other options? Like… I've seen some icons demi-sun-demi-moon. Not sure, but just to clarify before a research.

PR #1160, opened 2022-07-26
Refactor RubyObject with dry-struct

colby-swandale:
## Overview

At the moment, `RubyObject`, `RubyMethod`, `RubyConstant` & `RubyAttribute` are Plain Ol' Ruby Objects, all of which need to initialize and manage their state individually, leading to bloated classes that are less focused on the critical bits we care about.

This PR refactors many core models based on [dry-types](https://dry-rb.org/gems/dry-struct/1.0/), allowing us to abstract away the shapes for `RubyObject` et al. into schemas.

This PR also adds in FactoryBot to better manage objects during tests.


Gemfile
```diff
@@ -37,6 +37,8 @@ gem "rbs", require: false
 gem "anyway_config"
 gem "slim"
 gem "redis-session-store"
+gem "dry-struct"
+gem "factory_bot"
```
natematykiewicz, line 41: Any reason you're including FactoryBot in the production environment? I've always seen this in the "development & test" group.

colby-swandale: No reason, I will update the Gemfile 👍🏻

config/application.rb
```diff
@@ -18,6 +18,8 @@
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
 
+require_relative '../lib/types'
```
natematykiewicz, line 21: A common paradigm I've seen is to put files like this into `app/lib` instead of `lib`. You're breaking code reloading when you manually require like this. I'd suggest you move the folder and delete this line.

natematykiewicz:
I've never used any of the DRY gems before, but I do love this:
Screen Shot 2022-07-26 at 9 19 19 AM

test/test_helper.rb
```diff
@@ -6,10 +6,13 @@
 require "webmock/minitest"
 
 Dir.glob(Rails.root.join("lib/*.rb")).each { |f| require_relative f }
+Dir.glob(Rails.root.join("test/factories/*.rb")).each { |f| require_relative f }
 
 WebMock.disable!
 
 class ActiveSupport::TestCase
+  include FactoryBot::Syntax::Methods
+
```
natematykiewicz, line 15: You're not actually using this. Including this module allows you to remove `FactoryBot.` in your tests, so that `FactoryBot.build(:ruby_object, c: Integer)` becomes `build(:ruby_object, c: Integer)`. Since I like consistency, I feel like you should either remove the include, or remove `FactoryBot.` in the tests. Otherwise it feels like it leaves the door open for someone else to omit the `FactoryBot.` in other tests even though you're including it in these tests.

AlexWayfer (commented):
I don't see a lot of profit, but I'm not against.

natematykiewicz:
> I don't see a lot of profit, but I'm not against.

I find it to be much easier to read. Having a bunch of repetitive boilerplate code makes it harder to find the important parts. 

What's the real logic in here under the sea of:

```ruby
def {field_name}
  body[field_name]
end
```

Furthermore, using something to DRY this up helps make the code less error prone. It becomes harder to mismatch the method name and field on accident. Furthermore, the types on `dry-type` *could* be used to infer the GraphQL and ElasticSearch field types.

AlexWayfer:
@natematykiewicz maybe you're right. I could suggest `method_missing` or something like that, but there can be coercion, etc.

So, as I wrote, not a lot of profit, but some fair amount.

I've been working with some more high-level structures than just objects with attributes and can't judge them to the end.

config/application.rb
colby-swandale: Do you have any links to posts about this topic? I'm interested to learn more about this

natematykiewicz: https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#reloading
> Rails automatically reloads classes and modules if application files in the autoload paths change.

If it isn't autoloaded by Zeitwerk, it's not reloaded automatically when you write to the files. That means you need top stop and start `bin/dev` for those changes to take effect.

https://github.com/fxn/zeitwerk/issues/187#issuecomment-946179214
Xavier recommends `app/lib` instead of `lib` for any code that should be autoloadable (so like classes would go in app/lib and rake tasks go in lib).

colby-swandale: Apologies for the delay, moving `lib/types.rb` into `app/types` is better, thank you for sharing! 🙇🏻

test/test_helper.rb
colby-swandale: I agree, lets keep things consistent.

config/application.rb
natematykiewicz: No problem!

PR #1161, opened 2022-07-26
Revert "View Components"

colby-swandale:
Reverts rubyapi/rubyapi#1158


natematykiewicz:
What went wrong?

colby-swandale:
the page styles broke because i forgot to update the tailwind config 🤦🏻

PR #1233, opened 2022-10-12
Remove SLIM

colby-swandale:
This PR rewrites the remaining SLIM template files back to ERB and removes the Slim gem from the Gemfile.


app/views/home/index.html.erb
```diff
@@ -0,0 +1,40 @@
+<% description "Easily find and browse Ruby classes, modules and methods" %>
+<% content_for :head do %>
+  <script type="application/ld+json">
+      | {
```
natematykiewicz, line 4: This pipe is a typo, right? @colby-swandale

app/views/search/index.html.erb
```diff
@@ -0,0 +1,52 @@
...
+        <summary class="focus:outline-none">ElasticSearch Query</summary>
+        <pre><%= JSON.pretty_generate @search.query %></pre>
+      </details>
+    <% end %>
+  </div>
+  <% @search.results.each do |r| %>
+    <div class="md:p-3 py-3 my-1 rounded dark:text-gray-500">
+      <div class="flex">
+        <div class="w-full">
+          <% if r.is_a? RubyMethod %>
```
natematykiewicz, line 20: Since both halves of this are identical other than the method, it feels like this could get DRYed up with a little duck typing.

If a new method was added to all of possible types (name or something generic), then you could just call that method and it'd print the appropriate thing for the type.

colby-swandale: I agree with all your points, and honestly this whole page has been neglected and badly needs a bit of a refresh.

PR #1250, opened 2022-10-21
Fix styling for ordered lists

### LINKED ISSUE #1249: Ordered lists in documentation are not styled properly
Compare the documentation for the `Coverage` module on ruby-doc.org:

https://ruby-doc.org/stdlib-3.1.2/libdoc/coverage/rdoc/Coverage.html#module-Coverage-label-Usage

And on rubyapi.org:

https://rubyapi.org/3.1/o/coverage#module-Coverage-label-Usage

A couple other examples:
https://rubyapi.org/3.1/o/pp#method-c-width_for
https://rubyapi.org/3.1/o/readline#method-c-readline

It looks like there's a global CSS style that's interfering:

```css
menu, ol, ul {
  list-style: none;
  margin: 0;
  padding: 0;
}
```

Disabling `list-style: none` fixes the bug but also messes with the styling of the table of contents. Is it possible to only set `list-style: none` outside of the documentation?

cstyles:
Without this change, ordered lists inside documentation will have the `list-style: none` property which causes them to be displayed improperly. We're already overriding the style for unordered lists on the line above so let's do the same for ordered lists.

Fixes #1249

### DIFF
--- a/app/assets/css/application.css
+++ b/app/assets/css/application.css
@@ -70,6 +70,7 @@ pre:not(.ruby) {
 
 .ruby-documentation ul { @apply pl-4; }
 .ruby-documentation ul li { @apply list-disc; }
+.ruby-documentation ol li { @apply list-decimal; }
 .ruby-documentation dd, .ruby-documentation dt { @apply w-full text-center; }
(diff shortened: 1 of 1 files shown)


colby-swandale:
Thank you for the fix! 🙇🏻

PR #1287, opened 2022-11-27
Theme Dropdown

### LINKED ISSUE #1283: Theme selection does not appear to be respected
Following the merge of #1271, the site no longer appears to reflect my selected theme, instead always loading in light mode (with the exception of the main page at https://rubyapi.org, which loads in my system theme and does NOT have the theme selector). I can see the cookie being set correctly, however the page does not appear to reflect it, and the dropdown always shows the `System` theme as active no matter which option is actually chosen. Furthermore, despite `System` being selected, the light theme always seems to be served regardless of system preferences.

## Environment

- OS: macOS Ventura 13.0.1
- Browser: Mozilla Firefox 107.0, Google Chrome 107.0.5304.110

colby-swandale:
This PR is from #1271 (reverted) and addresses the problem described in #1283. The main problem was that the `system` theme was not respecting the `prefers-color-scheme` because we configured TailwindCSS to use the `dark` class toggle instead. 

This PR adds a new `meta` and `dynamic` attribute to `Theme`, which respectively sets the `<meta name=”theme-color”>` tag and loads `theme.js`, which is doing the same work as [essentials.js](https://github.com/rubyapi/rubyapi/blob/b1b4c647527884337ab0217a07bc68d0771d846b/app/javascript/essentials.js)

Closes #1283


natematykiewicz:
Looks great! I like how simple that meta tag was to add to `config/theme.yml` (and how clean that config is in general). I real need to read up on the `Vary` header to understand what you've done here. How is it possible to use a CDN to public cache something that depends on session?

PR #1294, opened 2022-12-01
Typescript

colby-swandale:
## Overview

I’ve often run into problems with our frontend code around types and what values stimulus, Monoco, etc, are expecting. We do not have unit tests (yet) for many Stimulus controllers, so it’s difficult to pick up on mistakes, typos, or general issues. This problem is an excellent opportunity for Typescript to help solve. We can take advantage of strict type-checking to help pickup on lots of these issues.

This PR moves all of our Front-end code from Javascript to Typescript. I've also added eslint to help ensure we're conforming to best practices, similar to Standard.


natematykiewicz:
As someone who thinks JavaScript is a garbage language, I'm really excited to see Typescript be added. 

JavaScript is unintuitive in many ways (especially regarding truthiness). Having your editor tell you that your arguments don't do what you think they do is really nice.

natematykiewicz:
Even with a proper test suite, I still think Typescript makes development much easier.

PR #1378, opened 2023-01-25
Ruby release Git Tag & branches information

### LINKED ISSUE #1359: Wrong link to GitHub code URL
1. Visit
https://rubyapi.org/3.2/o/enumerable#method-i-sort_by
2. Click “View on GitHub”
3. See a 404 at https://github.com/ruby/ruby/blob/v3_2/enum.c#L1472

It should be:
https://github.com/ruby/ruby/blob/v3_2_0/enum.c#L1472

Note the _0 in the URL.

colby-swandale:
This PR adds Git branch & tag information to the Ruby config to allow us to correctly build links and references to ruby-src code on GitHub.com.

Closes #1359

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -12,13 +12,9 @@ module ApplicationHelper
   def github_url(ruby_doc)
     version, file, line = ruby_doc.source_location.split(":")
 
-    if Current.ruby_version.dev?
-      # Not using URI.join, for a performance optimization. URI.join does A LOT of allocations.
-      # We know that our source_location is safe, because we make it in the importer.
-      # Inspiration: https://github.com/rack/rack/pull/1202
-      %(#{GITHUB_REPO}/master/#{file}#{"#L#{line}" if line})
-    else
-      %(#{GITHUB_REPO}/v#{version.tr(".", "_")}/#{file}#{"#L#{line}" if line})
-    end
+    # Not using URI.join, for a performance optimization. URI.join does A LOT of allocations.
+    # We know that our source_location is safe, because we make it in the importer.
+    # Inspiration: https://github.com/rack/rack/pull/1202
+    %(#{GITHUB_REPO}/#{Current.ruby_version.git_ref}/#{file}#{"#L#{line}" if line})
   end
 end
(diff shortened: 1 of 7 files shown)


app/models/ruby_version.rb
```diff
@@ -29,6 +30,18 @@ def prerelease?
     dev? || prerelease
   end
 
+  def git_ref
+    git_tag.present? ? git_tag : git_branch
```
natematykiewicz, line 34: I'm a big fan of `presence`, so that you don't have to reference `git_tag` twice.

https://github.com/rails/rails/blob/18707ab17fa492eb25ad2e8f9818a320dc20b823/activesupport/lib/active_support/core_ext/object/blank.rb#L45-L47

```suggestion
    git_tag.presence || git_branch
```

app/helpers/application_helper.rb
```diff
@@ -10,15 +10,11 @@ def homepage?
 
   # Map a method source file into a url to Github.com
   def github_url(ruby_doc)
-    version, file, line = ruby_doc.source_location.split(":")
+    _version, file, line = ruby_doc.source_location.split(":")
```
natematykiewicz, line 13: I worked on this a little a few days ago and ended up writing the exact same thing. It sure makes me wish that `ruby_doc` had a reference to the version that the method is for. Then this could actually be an instance method on `ruby_doc`.

Like:

```ruby
ruby_doc.github_url
```

config/ruby.yml
```diff
@@ -4,33 +4,59 @@ default: &default
       default: true
       sha256: cca9ddbc958431ff77f61948cb67afa569f01f99c9389d2bbedfa92986c9ef09
       url: https://cache.ruby-lang.org/pub/ruby/3.2/ruby-3.2.0.zip
+      git:
+        tag: v3_2_0
+        branch: ruby_3_2
```
natematykiewicz, line 9: I like this! I felt a bit stuck when looking into this because I had a single `git_ref` string might be a tag or might be a branch, and it felt really messy.

natematykiewicz (approved):

app/helpers/application_helper.rb
colby-swandale: I think this is a good direction to advocate for in the future!

PR #1410, opened 2023-02-21
Add podman as docker alternative

andyw8:
Due to changes in Docker's licensing, many companies no longer allow it to be used, but podman is a drop-in replacement in most cases.

(I had initially tried `podman compose` instead of `podman-compose`, so hopefully this will save someone else a little time).

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,11 @@ Start ElasticSearch:
 ```sh
 docker compose up -d
 ```
+or if you are using podman:
+
+```sh
+podman-compose up -d
+```
 
 Start the Rails Server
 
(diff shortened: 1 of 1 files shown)


colby-swandale:
I’m torn on adding this to the README. I’m totally fine with Podman and happy for people to use it, but I’m not keen to taking on support for container engines outside of Docker at the present moment. I’m worried that if we merge this and I change the `docker-compose.yml` tomorrow that causes Podman to break, am I now obligated to fix the issue?

Alternatively, there might be a middle ground where we put this into a collapsible section and include a warning to say Podman is not officially supported?

* https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-collapsed-sections

andyw8:
@colby-swandale That's fair, I've pushed an update.

PR #1426, opened 2023-03-04
Support `latest` alias

andyw8:
From the suggestion in https://github.com/rubyapi/rubyapi/pull/1417#issuecomment-1451696255


config/routes.rb
```diff
@@ -8,7 +8,7 @@
 
   ruby_versions = RubyConfig.ruby_versions.collect { |v| Regexp.escape(v.version) }
 
-  scope "(:version)", constraints: { version: /#{ruby_versions.join("|")}/ } do
+  scope "(:version)", constraints: { version: /#{ruby_versions.join("|")}|latest/ } do
```
andyw8, line 11: Since `RubyConfig.version_for("latest")` would return nil, no other change is needed.

https://github.com/rubyapi/rubyapi/blob/6fea025bf7626fedbf56d72489cb70d9154c4237/app/controllers/application_controller.rb#L8

andyw8:
Updates to use `current`.

natematykiewicz:
I've not pulled this up locally yet, but do all of the links display as "current" or "3.2" when you pull up "current" documentation?

andyw8:
As "3.2"

natematykiewicz:
That seems fine for now. Maybe later we can support making those say "current". So that your documentation links don't lose that.

For example here:
image

natematykiewicz (approved):
LGTM. Thoughts @colby-swandale?

config/routes.rb
colby-swandale: I would prefer to not rely on an unspecified rule as the way to support this feature. It's hard for anyone reading the code for the first time to understand this and it's very easy to break unintentionally.

cpjmcquillan:
This would be great for something like a [DuckDuckGo bang](https://duckduckgo.com/bangs) to search the docs.

FWIW, in that scenario it wouldn't be a problem for me personally if the links referred to the specific version 👍

PR #1844, opened 2024-01-26
Fix command typo in README

npras:
The import task is to be
run via `bin/rake`, not via
`bin/rails`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,13 +64,13 @@ bin/dev
 Ruby's documentation can be imported very easily. There's a rake task that will let you import a given versions' documentation:
 
 ```sh
-bin/rails import:ruby[3.1]
+bin/rake import:ruby[3.1]
 ```
 
 or you can easily import the latest versions of all currently supported versions of ruby:
 
 ```sh
-bin/rails import:ruby:all
+bin/rake import:ruby:all
 ```
 
 ## Running tests


natematykiewicz:
Rake tasks have been able to be invoked via `bin/rails` since Rails 5.0.

https://guides.rubyonrails.org/5_0_release_notes.html#railties-notable-changes
> Proxy Rake tasks through bin/rails. ([Pull Request](https://github.com/rails/rails/pull/22457), [Pull Request](https://github.com/rails/rails/pull/22288))

It was done to resolve [this issue](https://github.com/rails/rails/issues/18878).

Are you having a problem running them via `bin/rails`?

npras:
Oh, I thought that was only for the rake commands that came with rails (rake routes, rake db:migrate etc). Didn't think it worked for custom tasks as well. My bad.

I tried it now with `bin/rails import:ruby:all` and it's working just fine.

But the reason I opened this PR was because I got error when I tried these:

```
$ bin/rails import:ruby[3.1]
zsh: no matches found: import:ruby[3.1]

$ bin/rails import:ruby[3.3]
zsh: no matches found: import:ruby[3.3]

$ bin/rails import:ruby[3.0]
zsh: no matches found: import:ruby[3.0]

$ bin/rails import:ruby[2.5]
zsh: no matches found: import:ruby[2.5]

$ bin/rails import:ruby["3.3"]
zsh: no matches found: import:ruby[3.3]
```

But turns out, this is an issue with zsh. Escaping the brackets fixes this: `bin/rails import:ruby\[3.3\]`

Anyway, great project. I wanted to contribute, while I pick some new things myself. I'll go through code and send relevant PRs if needed.

natematykiewicz:
No worries! You know I've had that problem as well. The commands in the docs work fine in Bash, but as you said, zsh needs the brackets escaped.

Perhaps there's a better way to write this in the docs that works across shells better?

Maybe we should use Thor instead of Rake (like Rails does)?

Then we could do:

```bash
bin/rubyapi import 3.0
bin/rubyapi import:all
```

natematykiewicz:
@colby-swandale what do you think of switching the import tasks from Rake to Thor to avoid needing to escape the brackets? It'd also allow for some handy help commands.

natematykiewicz:
I'm not sure how to pull off `bin/rails import:ruby:3.2`. As far as I understand, Rake arguments have to be in brackets. Adding quotes to the documentations seems easy enough. I didn't realize that was an option. I thought it needed backslashes.

colby-swandale:
We would scrap the argument and generate a new task for each ruby version.

PR #2383, opened 2026-01-01
Fix segmentation fault in import:ruby task

TaoufikMejri:
## Problem
During the test-import check in PR https://github.com/rubyapi/rubyapi/pull/2361, I encountered a segmentation fault in the rbs gem at version 3.9.5.

> /home/runner/work/rubyapi/rubyapi/vendor/bundle/ruby/3.4.0/gems/rbs-3.9.5/lib/rbs/parser_aux.rb:28: [BUG] Segmentation fault at 0x00007fae6517fd60
> ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +YJIT +PRISM [x86_64-linux]

## Solution
Upgrade the rbs dependency from 3.9.5 to 3.10.0.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,6 +59,8 @@ jobs:
     env:
       SEARCH_URL: "http://localhost:9200"
       DATABASE_URL: "postgresql://rubyapi:rubyapi_password@localhost:5432/rubyapi_test"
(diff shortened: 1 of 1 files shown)


colby-swandale:
Thank you for helping improve Ruby API!

PR #2397, opened 2026-04-13
Fix canonical link url

4lllex:
The canonical URL was being generated with a `RubyRelease` id instead of its `version`. This resulted in incorrect canonical links, `https://rubyapi.org/1/o/string` instead of `https://rubyapi.org/4.0/o/string`

Screen Shot 2026-04-13 at 5 11 15 AM

The fix is to pass the actual version value:
```rb
version: Current.default_ruby_release.version
```

Alternatively, [`to_param`](https://api.rubyonrails.org/classes/ActiveRecord/Integration.html#method-i-to_param) method could be defined in the model to return the `version`.


colby-swandale (approved):
Thanks!

PR #2398, opened 2026-04-13
Fix local development setup

### LINKED ISSUE #2396: Local/dev setup is broken
Local setup instructions don't work: https://github.com/rubyapi/rubyapi#local-machine. Database never gets connected.

https://github.com/rubyapi/rubyapi/blob/06052ae72281be05a4065b97cd87e10a5cc210cc/config/database.yml#L22-L26

Is `DB_HOST` env needs to be set up? without it, database is trying to connect to a local socket. Even if `DB_HOST` is set to `localhost` username and password don't match the local db:

https://github.com/rubyapi/rubyapi/blob/06052ae72281be05a4065b97cd87e10a5cc210cc/compose.yml#L45-L46

---

Healthcheck keeps checking forever as it is trying to connect to `rubyapi` database which doesn't exist:

https://github.com/rubyapi/rubyapi/blob/06052ae72281be05a4065b97cd87e10a5cc210cc/compose.yml#L51

4lllex:
This PR fixes issues with the local development environment setup as documented in the readme:

```sh
bin/setup 
docker compose up
bin/dev
```

The two main issues:
 - rails cannot connect to Postgres
 - docker services never become healthy

Changes:
- fix the OpenSearch healthcheck to use port `9200`
- accept OpenSearch cluster status yellow for healthchecks, it is expected for local single node
- make the database port configurable via `DB_PORT` for both Docker and Rails. 
- add the database name to the Postgres healthcheck with `-d rubyapi_development`
- remove committed `database.yml`; it is gitignored and should be generated by `bin/setup`
- update `config/database.yml.sample` to match the Docker Compose configuration

Fixes #2396

Perhaps removing `database.yml` was incorrect https://github.com/rubyapi/rubyapi/commit/8d7decf3ba0076cc4c1222324d1168f3706cb266, but also conflicting instructions: 
https://github.com/rubyapi/rubyapi/blob/06052ae72281be05a4065b97cd87e10a5cc210cc/.gitignore#L50 https://github.com/rubyapi/rubyapi/blob/06052ae72281be05a4065b97cd87e10a5cc210cc/bin/setup#L20


bin/setup
```diff
@@ -25,10 +25,4 @@ FileUtils.chdir APP_ROOT do
 
   puts "\n== Removing old logs and tempfiles =="
   system! 'bin/rails log:clear tmp:clear'
-
-  unless ARGV.include?('--skip-server')
```
colby-swandale, line 29: Let's stay with what rails gives us out the box

config/database.yml.sample
```diff
@@ -19,11 +19,10 @@ default: &default
   # https://guides.rubyonrails.org/configuring.html#database-pooling
   max_connections: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
   gssencmode: <%= RUBY_PLATFORM =~ /darwin/ ? "disable" : "prefer" %>
-  <% if ENV["DB_HOST"] %>
-  host: <%= ENV["DB_HOST"] %>
-  username: postgres
-  password: postgres
-  <% end %>
+  host: <%= ENV.fetch("DB_HOST") { "localhost" } %>
```
colby-swandale: I would also prefer sticking to what rails gives us out the box here as well. If you need to specific configuration, then just update the `database.yml`

compose.yml
```diff
@@ -24,12 +24,13 @@ services:
     ports:
       - "5601:5601"
     environment:
-      OPENSEARCH_HOSTS: '["https://search:9200"]'
```
colby-swandale, line 27: Production runs against AWS OpenSearch with real TLS verification on the client side. If dev drops to plain HTTP, the client's TLS config (verify mode, CA bundle, cert handling) isn't exercised until we deploy. I'd rather keep HTTPS locally so misconfigurations surface here, not in production

config/database.yml.sample
colby-swandale:
```suggestion
  host: <%= ENV.fetch("DB_HOST") { "127.0.0.1" } %>
```

compose.yml
4lllex: I had to remove `plugins.security.disabled=true` and update the `opensearch` healthcheck to get the console working, otherwise it failed to connect with:

```sh
search-console-1  | {"type":"log","@timestamp":"2026-05-02T05:29:09Z","tags":["error","opensearch","data"],"pid":1,"message":
  "[ConnectionError]: write EPROTO C0271D12407D0000:error:0A00010B:SSL routines:ssl3_get_record:wrong version number:../deps/openssl/openssl/ssl/record/ssl3_record.c:355:\n"}
```

This also requires 
```
OPENSEARCH_URL="https://admin:admin@127.0.0.1:9200"
```
set in `development`. Do you want to set it as a default somewhere?
https://github.com/rubyapi/rubyapi/blob/7c7fd9918ec99dd502b76cf6b75d87f5ce0bb4a5/config/initializers/searchkick.rb#L1

colby-swandale: This is a good opportunity for `SearchConfig` to take back managing `OPENSEARCH_URL`

colby-swandale (approved):
Tested this locally and got the app running without any issues. Thanks!

4lllex:
Without setting `username:password` searchkick cannot connect:

```sh
$ OPENSEARCH_URL="https://localhost:9200" bin/rails runner "p RubyObject.reindex"
The client is unable to verify distribution due to security privileges on the server side. Some functionality may not be compatible if the server is running an unsupported product.
/home/alex/.rbenv/versions/4.0.3/lib/ruby/gems/4.0.0/gems/opensearch-ruby-3.4.0/lib/opensearch/transport/transport/base.rb:227:in 'OpenSearch::Transport::Transport::Base#__raise_transport_error':
  [401]  (OpenSearch::Transport::Transport::Errors::Unauthorized)
```
```sh
$ OPENSEARCH_URL="https://admin:admin@localhost:9200" bin/rails runner "p RubyObject.reindex"
true
```

https://github.com/rubyapi/rubyapi/blob/685b4a2d322b309a1a3e8dc84512bb9117df35a5/config/opensearch.yml#L1-L5

colby-swandale:
Followed up in #2432

PR #2400, opened 2026-04-15
Fix missing class/module label

### LINKED ISSUE #2395: Missing class/module labels
This section doesn't render for modules or classes:

https://github.com/rubyapi/rubyapi/blob/06052ae72281be05a4065b97cd87e10a5cc210cc/app/views/objects/show.html.erb#L245-L253
Example, https://rubyapi.org/4.0/o/string, there should be a `Class` label on the right side of the `String`.

4lllex:
Fix predicate checks for `object_type`: https://github.com/rubyapi/rubyapi/blob/06052ae72281be05a4065b97cd87e10a5cc210cc/app/models/ruby_object.rb#L75-L81 to match stored values (`"class_object"`, `"module_object"`): https://github.com/rubyapi/rubyapi/blob/06052ae72281be05a4065b97cd87e10a5cc210cc/lib/rubyapi_rdoc_generator.rb#L96

Update fixtures and add a test for label rendering.

Before  Screenshot from 2026-04-14 20-47-45
After  Screenshot from 2026-04-14 20-47-52

Fixes #2395

### DIFF
--- a/app/models/ruby_object.rb
+++ b/app/models/ruby_object.rb
@@ -73,11 +73,11 @@ class RubyObject < ApplicationRecord
   end
 
   def class_object?
-    object_type == "class"
+    object_type == "class_object"
   end
 
   def module_object?
-    object_type == "module"
+    object_type == "module_object"
   end
 
   def depth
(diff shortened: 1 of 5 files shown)


colby-swandale (approved):
Thanks!

PR #2433, opened 2026-05-03
Put Docker setup first in README

4lllex:
Can't prepare the database if it is not up. Update readme to run `docker compose up -d` before `bin/setup`.


colby-swandale (approved):
Thanks for the contribution!

PR #2470, opened 2026-09-17
Limit Tailwind watch to app, lib, and config

4lllex:
Don't watch `tmp` and `log`. Reduced measured watches from 16,000+ to 29 and stopped rebuilding after and during every request on log writes.

Before:
```sh
$ rg "^inotify" /proc/$(pgrep -nf "tailwindcss.*rubyapi")/fdinfo/* | wc -l
16070
```

After:
```sh
$ rg "^inotify" /proc/$(pgrep -nf "tailwindcss.*rubyapi")/fdinfo/* | wc -l
29
```


colby-swandale:
I appreciate you looking into the problem, but this adds quite complex configuration for tailwind than I’m keen to see

natematykiewicz:
@colby-swandale Tailwind 3 was configured to only watch:

```js
  content: [
    './app/views/**/*.html.(slim|erb)',
    './app/helpers/**/*.rb',
    './app/javascript/**/*.js',
    './app/components/**/*.html.(slim|erb)',
  ],
```

It seems like the [upgrade to Tailwind 4](https://github.com/rubyapi/rubyapi/commit/98ae01f120a1c5870c269aa533ffac3a5bc3c378#diff-ddc1067f080930bc48865f4660316a14b21bbc9f9ba85975ce8479342b983d2eL3-L6) (tailwind.config.js) caused it to watch the entire repo.

What's the Tailwind 4 way to make the watcher only watch relevant directories?

